PR #2, opened 2016-02-25
Fast function map

LMescheder:
This pull request makes linear mappings type stable in julia 0.5 and when using the [Fast Anonymous](https://github.com/timholy/FastAnonymous.jl) package.
In order to do so, it splits the current `FunctionMap` type into `FunctionMap` and `MutatingFunctionMap` with common parent `AbstractFunctionMap`. This allows to use dispatch to chose the right multiplication method.
Contrary to the current solution, `issym` and `ishermitian` is only employed when `ftranspose` and `ctranspose` are not set. This was necessary to enforce type stability, but I believe it is also quite sensible.

### DIFF
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -62,10 +62,22 @@ include("functionmap.jl") # using a function as linear map
 
 LinearMap{T}(A::Union{AbstractMatrix{T},AbstractLinearMap{T}};isreal::Bool=Base.isreal(A),issym::Bool=Base.issym(A),ishermitian::Bool=Base.ishermitian(A),isposdef::Bool=Base.isposdef(A)) =
     WrappedMap(A;isreal=isreal,issym=issym,ishermitian=ishermitian,isposdef=isposdef)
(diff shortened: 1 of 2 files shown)


Jutho:
Thanks for this. I'll try to make some time to review as soon as possible. LinearMaps.jl has not received any major update since its initial release which was written for julia 0.3, only a small compatibility update for 0.4. So this is very welcome.

Jutho:
I am sorry but this has slipped my mind. I did some updates to LinearMaps.jl (master) today, to make it compatible with v0.5. The biggest change involves replacing `issym` with `issymmetric` everywhere. Would you mind rebasing this PR?

ChrisRackauckas:
FastAnonymous shouldn't be necessary with v0.5 since anonymous functions are generic. Since v0.5 is now required, should this be closed?

Jutho:
I think LinearMaps.jl has received updates that are in line with the idea of this PR. Unless there are further questions or suggestions, I will close this PR.

PR #8, opened 2017-01-07
Fallback options when complement function is not available

shivin9:
@Jutho A complement function is not always available, so in case we want to find out the inverse mapping then that is done by multiplying the linear map's complement with the vector. ie'`full(L)'*x`

### DIFF
--- a/src/functionmap.jl
+++ b/src/functionmap.jl
@@ -16,6 +16,11 @@ FunctionMap{T}(f, fc, M::Int, ::Type{T} = Float64; kwargs...) = FunctionMap(f, f
 
 function FunctionMap{T,F1,F2}(f::F1, fc::F2, M::Int, N::Int, ::Type{T} = Float64;
     ismutating::Bool = false, issymmetric::Bool = false, ishermitian::Bool=(T<:Real && issymmetric), isposdef::Bool = false)
+
+    if fc == nothing
+        warn("transpose not implemented for the map")
+    end
+
     FunctionMap{T,F1,F2}(f, fc, M, N, ismutating, issymmetric, ishermitian, isposdef)
 end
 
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Can this get reviewed? Something like this is necessary for full usage with IterativeSolvers.jl

Jutho:
Is this really a good strategy? For large problems, generating the full matrix might just be to memory costly. Furthermore, with this approach, it is regenerated for every single application of At_mul_B and friends, which is highly inefficient.

I think it is better to error out in this case, so that the user knows he explicitly also has to define a transpose function, in order to be able to use it with e.g. an iterative solver that requires the transpose action. Any thoughts?

shivin9:
Well many a times it's not that straight forward to write the transpose function and if the user wants to calculate the complement anyhow, then it can helpful if we give him an option. They are used quite frequently in iterative algorithms.
I will optimize it out soon btw.

Jutho:
I still think this is not the way to go. In particular, you now always allocate a matrix zeros(T,M,N) when `fc` is not provided. This is too costly; one of the main points of `LinearMaps` is to represent operators which are too costly to store as a matrix.

Jutho:
As a function? I wouldn't know how to do it. For the solution you were using above, I think you can just replace `LinearMap(f)` with `full(LinearMap(f))` wherever you are using it. Then the linear solver will be able to use the transpose, just because it has the full matrix.

ChrisRackauckas:
That's a good point. So it's probably better to let the user `full` and just let it error if no complement is given. If there was a `sparse` function that would probably be the best option. Of course, if there's a way to construct the transposed operation that would be ideal, but whether it's possible eludes me.

PR #10, opened 2017-04-27
Add new harmless LinearMap properties

lopezm94:
Sometimes when using matrices we do harmless tests with `ndims` and `length` in IterativeSolvers, this could help abstract LinearMaps and only discovering its type when strictly necessary.


Jutho:
These can probably be implemented at the level of 'AbstractLinearMap'?

lopezm94:
Yes, you are right. I moved it to the abstract and added some tests.

PR #12, opened 2017-05-16
Updates for v0.6 compatibility

ScottPJones:


Jutho:
Thanks for this, I was actually working on a similar update for v0.6.0 (see branch called v1.0) which will partially overlap with this PR, and partially conflict. So I'll hold off merging for now if that's ok and try to  finish my own take asap.

ScottPJones:
Ah!  I checked to see if anybody else had a PR before I proceeded, but didn't look for stuff on other branches.
I was trying to 1) get rid of deprecation warnings in DifferentialEquations 2) make sure any packages used got tested on v0.6 (and latest) while 3) retaining (at least) v0.5 compatibility.
I see some of your changes on the v1.0 branch are only for v0.6, and that you are dropping v0.5.x support?
Are you interested in any of the other changes I made (adding v0.6 / nightly tests, allowing errors for now in the v0.6 testing [until it is released], badges for v0.5/v0.6), in a few places making the code not go over the 92 char limit recommended by the Julia contribution guide)?  If so, I'll update my PR as soon as you've merged your changes.

Jutho:
My reason for dropping v0.5.x support was pure laziness, as well as not wanting to learn to use Compat. You seem to have some nice tricks to avoid using Compat. I will certainly study your PR, and either include these ideas directly or merge them afterward. Also the updated CI and badges will definitely be included.

Jutho:
I have since updated LinearMaps.jl with a new version that is compatible with julia v0.6.0, and in fact no longer with older versions. I did in the end decide to drop support for older versions because of some additional changes that I made. Unless you have further suggestions I believe this PR can be closed?

PR #16, opened 2017-10-12
Update linearcombination.jl

raminammour:
Right now, this is an error:

```julia
L=LinearMap(x->rand(10),x->rand(10),10)
B=2*L
B'*rand(10)

ERROR: MethodError: no method matching Ac_mul_B(::LinearMaps.LinearCombination{Float64,Tuple{LinearMaps.FunctionMap{Float64,##33#35,##34#36}},Tuple{Int64}}, ::Array{Float64,1})
```
The code above fixes it.

### DIFF
--- a/src/linearcombination.jl
+++ b/src/linearcombination.jl
@@ -126,3 +126,28 @@ function Base.Ac_mul_B!(y::AbstractVector, A::LinearCombination, x::AbstractVect
     end
     return y
 end
+
+function Base.At_mul_B( A::LinearCombination, x::AbstractVector)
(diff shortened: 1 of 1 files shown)


Jutho:
This should be covered by the fallback methods of the `LinearMap` type, but something was broken indeed. I fixed this differently in the most recent commit, and as soon as CI turns green, I'll tag a new version

Jutho:
Does the new version (which is now tagged and registered) solve your issue?

raminammour:
That fixes it, thank you!

Sorry for the late reply, I had no access to the code for the last few days.

PR #17, opened 2017-11-28
Construct sparse matrix from LinearMap

samuelpowell:
Could pre-allocate a vector for use constructing `Js` if that would be preferable.

### DIFF
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -88,6 +88,26 @@ function Base.full(A::LinearMap)
     return mat
 end
 
+# sparse: create sparse matrix representtion of LinearMap
+function Base.sparse(A::LinearMap)
+    M, N = size(A)
+    T = eltype(A)
+    Is = Int[]
+    Js = Int[]
+    Vs = T[]
+    v = zeros(T, N)
+    for i = 1:N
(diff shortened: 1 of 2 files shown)


Jutho:
Looks good. As I never use sparse arrays (in the traditional sense), I haven't been following discussions in Julia Base closely. Do you know what the faith of sparse arrays and the corresponding constructors will be? Will `Base.sparse` still be a method in the upcoming 0.7/1.0-prerelease ?

samuelpowell:
Hi @Jutho, The `Base.sparse` constructor used here is fairly fundamental (takes the `COO` form and converts to `CSC` format), and I can see no evidence of any pending changes. The deprecation of `full` et al. will need changes to the test (but then that applies to the dense array methods as well).

Having thought about this I could actually use the inner `SparseMatrixCSC` which would be more efficient (and is as low level as it gets). Will push some changes shortly.

samuelpowell:
I have updated to build the CSC vectors directly. This is faster and involves fewer allocations. Added a test where there is actually some sparsity!

Jutho:
Thanks for this!

samuelpowell:
My pleasure, an extremely useful package @Jutho.

PR #23, opened 2018-06-27
Julia 0.7 upgrade

dkarrasch:


README.md
```diff
@@ -88,13 +92,13 @@ None of the types below need to be constructed directly; they arise from perform
...
 
-*   `LinearCombination`, `CompositeMap`, `TransposeMap` and `CTransposeMap`
+*   `LinearCombination`, `CompositeMap`, `TransposeMap` and `AdjointMap`
 
     Used to add and multiply `LinearMap` objects, don't need to be constructed explicitly.
 
 ## Examples
 
-The `LinearMap` object combines well with the iterative eigensolver `eigs`, which is the Julia wrapper for Arpack.
+The `LinearMap` object combines well with the iterative eigensolver `eigs`, which is the Julia wrapper for Arpack, and with iterative solvers from [IterativeSolvers.jl](https://github.com/JuliaMath/IterativeSolvers.jl).
```
Jutho: This should now probably mention to `[Arpack](https://github.com/JuliaLinearAlgebra/Arpack.jl)`

src/LinearMaps.jl
```diff
@@ -3,14 +3,13 @@ module LinearMaps
...
-if VERSION >= v"0.7.0-DEV.1415"
-    const adjoint = Base.adjoint
-else
-    const adjoint = Base.ctranspose
-end
+import LinearAlgebra, SparseArrays
 
+import Base: +, -, *, \, /, ==
+import LinearAlgebra: transpose, adjoint
+import LinearAlgebra: issymmetric, ishermitian, isposdef
```
Jutho: Personally, I am more a fan of using fully qualified names when overloading methods (see below) and thus, not importing all of them. I did seem to make an exception for operators in the old version, but even there `Base.:(+)` is not too bad.

src/LinearMaps.jl
```diff
@@ -3,14 +3,13 @@ module LinearMaps
...
+import LinearAlgebra: BLAS.axpy!, mul!, lmul!, UniformScaling
+import SparseArrays: sparse
```
Jutho: I don't think we provide any new method definition for `UniformScaling`, so do we need that one?

src/LinearMaps.jl
```diff
@@ -20,87 +19,55 @@ Base.eltype(::LinearMap{T}) where {T} = T
 Base.eltype(::Type{L}) where {T,L<:LinearMap{T}} = T
 
 Base.isreal(A::LinearMap) = eltype(A) <: Real
-Base.issymmetric(::LinearMap) = false # default assumptions
-Base.ishermitian(A::LinearMap{<:Real}) = issymmetric(A)
-Base.ishermitian(::LinearMap) = false # default assumptions
-Base.isposdef(::LinearMap) = false # default assumptions
+issymmetric(::LinearMap) = false # default assumptions
```
Jutho: I would replace these with `LinearAlgebra.issymmetric(::LinearMap) = false` etc.

src/LinearMaps.jl
```diff
@@ -20,87 +19,55 @@ Base.eltype(::LinearMap{T}) where {T} = T
...
 Base.length(A::LinearMap) = size(A)[1] * size(A)[2]
 
-# any LinearMap subtype will have to overwrite at least one of the two following methods to avoid running in circles
-*(A::LinearMap, x::AbstractVector) = Base.A_mul_B!(similar(x, promote_type(eltype(A),eltype(x)), size(A,1)), A, x)
-Base.A_mul_B!(y::AbstractVector, A::LinearMap, x::AbstractVector) = begin
-    length(y) == size(A,1) || throw(DimensionMismatch("A_mul_B!"))
-    copy!(y, A*x)
+*(A::LinearMap, x::AbstractVector) = mul!(similar(x, promote_type(eltype(A),eltype(x)), size(A,1)), A, x)
+mul!(y::AbstractVector, A::LinearMap, x::AbstractVector) = begin
+    length(y) == size(A,1) || throw(DimensionMismatch("mul!"))
```
Jutho: Similarly here: `Base.:(+)` (possibly) and `LinearAlgebra.mul!` (preferably). 

Also, I don't know why I choose to write the `mul!` function as
```
function mul!( ) = begin
... (multiple lines)
end
```
instead of just
```
function mul!( )
... multiple lines
end
```

src/LinearMaps.jl
```diff
@@ -20,87 +19,55 @@ Base.eltype(::LinearMap{T}) where {T} = T
...
+At_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, transpose(A), x)
+At_mul_B(A, x) = transpose(A) * x
+Ac_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, adjoint(A), x)
+Ac_mul_B(A, x) = adjoint(A) * x
 
-# full: create matrix representation of LinearMap
-function Base.full(A::LinearMap)
+# Array: create matrix representation of LinearMap (shall we call it Base.Matrix?)
+function Base.Array(A::LinearMap)
     M, N = size(A)
```
Jutho: I am certainly fine with calling it `Base.Matrix`. Or maybe provide both.
```
Base.Array(A::LinearMap) = Matrix(A)
```

src/composition.jl
```diff
@@ -3,7 +3,7 @@ struct CompositeMap{T,As<:Tuple{Vararg{LinearMap}}} <: LinearMap{T}
     function CompositeMap{T,As}(maps::As) where {T,As}
         N = length(maps)
         for n = 2:N
-            size(maps[n],2)==size(maps[n-1],1) || throw(DimensionMismatch("CompositeMap"))
+            size(maps[n],2) == size(maps[n-1],1) || throw(DimensionMismatch("CompositeMap"))
         end
```
Jutho: 👍

src/composition.jl
```diff
@@ -73,72 +73,72 @@ end
 
 # special transposition behavior
 transpose(A::CompositeMap{T}) where {T} = CompositeMap{T}(map(transpose, reverse(A.maps)))
-adjoint(A::CompositeMap{T}) where {T} = CompositeMap{T}(map(adjoint, reverse(A.maps)))
+adjoint(A::CompositeMap{T}) where {T}   = CompositeMap{T}(map(adjoint, reverse(A.maps)))
```
Jutho: 👎  :)

Jutho: Ah now I see, you want both lines to be outlined at the `=` sign. Fine by me

src/transpose.jl
```diff
@@ -4,54 +4,51 @@ end
 struct AdjointMap{T, A<:LinearMap{T}} <: LinearMap{T}
     lmap::A
 end
-(::Type{TransposeMap})(lmap::LinearMap{T}) where {T} = TransposeMap{T, typeof(lmap)}(lmap)
-(::Type{AdjointMap})(lmap::LinearMap{T}) where {T} = AdjointMap{T, typeof(lmap)}(lmap)
+
+TransposeMap(lmap::LinearMap{T}) where {T} = TransposeMap{T, typeof(lmap)}(lmap)
+AdjointMap(lmap::LinearMap{T}) where {T}   = AdjointMap{T, typeof(lmap)}(lmap)
```
Jutho: I don't know if these methods are still necessary; they probably date back to from before triangular dispatch existed.

src/LinearMaps.jl
dkarrasch: I thought one needs to import those explicitly if one wants to overload them. I didn't mean to omit full qualifiers, that was more like ignorance. I'll certainly add the full qualifiers in the code and check if it is necessary to import overloaded functions here.

Jutho: If I understood correctly, and nothing has changed.
* If you only want to use a function (or anything else) from a package (including standard library packages), you do
    *   `using Package` if the function/object/name is exported
    *  `using Package` followed by `using Package: unexported_name` if the name is not exported. Or just `using Package` and then use the fully qualified name `Package.unexported_name`.

* If you want to overload a function (exported or not), you should:
    *   `using Package` followed by `import Package: name`
    *  or `using Package` and then just use the fully qualified name `Package.name`.

Jutho: You might want `using LinearAlgebra`, unless you also want to use fully qualified names whenever you just want to use/call one of these functions.

dkarrasch:
One more thing that I'm not sure what to do. I thought I read that one has to list all loaded packages in REQUIRE. If I simply add `LinearAlgebra` and `SparseArrays`, then it errors and says:
```
ERROR: The following package names could not be resolved:
 * LinearAlgebra (not found in project, manifest or registry)
 * SparseArrays (not found in project, manifest or registry)
Please specify by known `name=uuid`.
```
Probably we don't need to do anything so far, so I'll leave the one line `Julia 0.7-` in REQUIRE, unless you're certain about how to be more consistent.

src/composition.jl
dkarrasch: Github interprets your thumb as a change request, so I'll pretend I addressed the change request by this stupid answer. :-D

src/LinearMaps.jl
```diff
@@ -1,116 +1,76 @@
...
-
-if VERSION >= v"0.7.0-DEV.1415"
-    const adjoint = Base.adjoint
-else
-    const adjoint = Base.ctranspose
-end
+using LinearAlgebra
+import SparseArrays
 
+import Base: +, -, *, \, /, ==
```
Jutho: You can probably get rid of this line as well now, since you fully qualified all identifiers including operators.

dkarrasch: Just removing the line gives an error, and Julia is suggesting `You may have intended to import Base.*`. I'll try to `using Base` then...? The error occurs when I want to `A * v` in `sparse`.

Jutho: Maybe there is still some method definition for `*` somewhere that we've overlooked.

Jutho: I am fine merging like this. This looks great, and is by far the most polished PR I've ever seen. I could still look into this particular point once I have it here locally, but it is not important.

dkarrasch: The issue is that it errors when we use the symbols on the rhs, not when we are defining them. I think the function definitions are fine. It complains about line 62:
```
MethodError: no method matching *(::Array{Complex{Float64},2}, ::Array{Complex{Float64},1})
```
which sounds very weird. I can get rid of the error by replacing `A * v` by `Lv = Base.:(*)(A, v)`, but then it complains about the use of `\`. I guess one would have to fully qualify all operators throughout. Strangely, just `using Base` does not solve the issue, in contrast to `using LinearAlgebra`, where we are doing the same things: both using and extending functions.

dkarrasch: You're welcome. Was my pleasure, great package!

Jutho:
@dkarrasch , there was indeed one more definition of plain `*` instead of `Base.:(*)` on line 28 in wrappedmap.jl. But I think this definition can go all together.

I am very grateful for this PR. Would you be interested in becoming a collaborator in this package. I don't use it myself very often anymore, so it could certainly benefit from a more active maintainer. I have also expressed interest in transferring ownership to e.g. the `JuliaLinearAlgebra` group @JuliaLinearAlgebra , but I am not sure if they are interested in this.

dkarrasch:
Sorry I missed that. :-)

Yes, I'd be interested in contributing in maintaining this package. If the LinearAlgebra group is willing to take over, that would be also perfect. I do think they should have interest in that, given the `IterativeSolvers` package, which is (supposed to be/-come) a Julian Arpack replacement for  `eigs` and `svds`, and the very close connections established there. So, either way is fine for me.

Jutho:
@dkarrasch , I've invited you to become a collaborator, which I assume means you will have full admin privileges (I have not used this functionality of github before). In time, it would still be great if this package gets adopted by @JuliaLinearAlgebra or some related Julia group.

PR #27, opened 2018-07-11
Added mul!'s, convert's, tests; expanded README

dkarrasch:
* added mul! for application to matrices interpreted as blocks of vectors
* added a 5-args-mul! for TSVD.jl
* added convert to allow for Arpack.svds calls on LinearMap's
* added tests for all introduced functions, improved code style in tests
* documented all new functions in README


dkarrasch:
I don't understand why all the old commits are listed again... I thought I made only one commit to the new branch. Sorry once more.

dkarrasch:
@Jutho : Needs more testing to avoid allocations in the matrix-`mul!`. I felt that the `Matrix` trick would allocate first and then write inplace to `Y`. So I started to appreciate your loop proposal, i.e., applying `mul!` columnwise. Strangely,
```
@inbounds @views for i=1:size(X, 2)
        mul!(Y[:, i], A, X[:, i])
    end
```
allocates independently from the length of the columns and linearly with the size of the loop. The function call in each loop should be allocation-free, as tested for `mul!(::Vector,::LinearMap,::Vector)`. So I need to find out what is going on here.

Jutho:
Is the allocation large? The construction of a view allocates a little bit (I don't know how much exactly), so the number of views created is twice the size of the loop.

dkarrasch:
Yes, I think it's the two views. It's 96 bytes per loop iteration. I wasn't sure to which extent such things can be optimized. I guess that's acceptable, and it is much better than the simple `Matrix(A*X)` thing. Cool, that seems settled then.

There remains a second topic for optimization. Andreas Noack pointed me to his TSVD.jl package. The manual states that it requires a `mul!(\alpha,A,x,\beta,y)` implementation for the type of `A`, even though it seemed to work out of the box for `LinearMap`s (on Julia 0.6.3). It seems that we can't get rid of an allocation of the size of `y` there. If that's acceptable as well, then we have (soon) three SVD methods available: Arpack (works on Julia 0.7 with the new convert commands), TSVD and IterativeSolvers (both worked already on Julia 0.6.3)! I'll update the PR, which is then worth being reviewed.

src/LinearMaps.jl
```diff
@@ -25,16 +25,22 @@ function LinearAlgebra.mul!(y::AbstractVector, A::LinearMap, x::AbstractVector)
...
+    @inbounds @views for i=1:size(X, 2)
+        mul!(Y[:, i], A, X[:, i])
+    end
     return Y
 end
 # the following is required for TSVD.jl
-function LinearAlgebra.mul!(α::Number, A::LinearMap, x::Vector, β::Number, y::Vector)
-    y .= α * A * x + β * y
+function LinearAlgebra.mul!(α::Number, A::LinearMap, x::AbstractVector, β::Number, y::AbstractVector)
+    if β != one(β)
```
Jutho: The only precedent for a 5-argument `mul!` in Base/Stdlib (which I would very much like to see) seems to be for Sparse arrays (see https://discourse.julialang.org/t/why-doesnt-mul-have-a-gemv-style-interface/12136 ). It seems that, with Julia 0.7 and the transition from `A(x)_mul_B!` to `mul!`, the argument order has changed in a good way, no longer following BLAS but such that 3-arg mul is like 5-arg mul with default values for arguments: `mul!(C, A, B, alpha = 1, beta = 0)`. So I hope that `TSVD.jl` will follow that convention.

dkarrasch: Thanks @Jutho for your comments and the link. There were so many instances of `mul!` or their predecessors in Julia 0.6 that it was hard to find the 5-arg versions in the first place. As for Julia 0.7, I don't have Juno working with it, and I had no idea how to crawl the docs for the right argument signature (the julialang.org docs don't mention the 5-arg version). If there is a `mul!` with the `C, A, B` signature, I'll adopt to that. I'm sure Andreas Noack is aware of that convention and will follow. I'll file an issue at `TSVD.jl` to track any changes in a Julia 0.7 update. After that, I'd just merge (will need to ask my colleague about how to tag a new version etc.). Since this is an extension without breaking anything, this should be then version 2.0.1?

dkarrasch:
Aha, the 5-args mul! overwrote the 3-args mul! because of the default values for the scalars. I removed the 3-args one. Hopefully, the compiler realizes that the function with default values reduces to exactly the original 3-args code.

src/LinearMaps.jl
```diff
@@ -21,12 +21,40 @@ Base.size(A::LinearMap, n) = (n==1 || n==2 ? size(A)[n] : error("LinearMap objec
 Base.length(A::LinearMap) = size(A)[1] * size(A)[2]
 
 Base.:(*)(A::LinearMap, x::AbstractVector) = mul!(similar(x, promote_type(eltype(A), eltype(x)), size(A, 1)), A, x)
-function LinearAlgebra.mul!(y::AbstractVector, A::LinearMap, x::AbstractVector)
+function LinearAlgebra.mul!(y::AbstractVector, A::LinearMap, x::AbstractVector, α::Number=1, β::Number=0)
```
Jutho: Probably the compiler optimizes this away in most cases, but maybe change to:
```julia
function LinearAlgebra.mul!(y::AbstractVector, A::LinearMap{T}, x::AbstractVector, α::Number=one(T), β::Number=zero(T)) where {T}
```

dkarrasch: 👍

src/LinearMaps.jl
```diff
@@ -21,12 +21,40 @@ Base.size(A::LinearMap, n) = (n==1 || n==2 ? size(A)[n] : error("LinearMap objec
...
+        rmul!(y, β)
+        y .+= A * x
+        return y
+    elseif α == zero(α)
+        β == zero(β) && (fill!(y, zero(eltype(y))); return y)
+        β == one(β) && return y
+        rmul!(y, β)
+        return y
+    else # α != 0, 1
+        rmul!(x, α)
```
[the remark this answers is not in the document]
dkarrasch: Right, I was too focused on avoiding allocations again. I'll replace the else-branch (alpha != 0, 1) by
```
β == zero(β) && (A_mul_B!(y, A, x); rmul!(y, α); return y)
β == one(β) && (y .+= rmul!(A * x, α); return y)
# β != 0, 1
rmul!(y, β)
y .+= rmul!(A * x, α)
return y
```
that seems allocation-free for the beta==0 case. Otherwise, we have a single allocation of the size of `y`.

dkarrasch:
I guess we eliminated all controversial aspects and found good solutions, so I'd be happy to merge this. Since this is my first time, would "Squash and merge" be the right option from the dropdown options? Somehow, I want to get rid of all my previous commits, and "start from scratch" for new PRs.

Also, for people to be able to use the matrix-based `mul!`, we would need to tag a new version, right? I'd also be happy to do this, so the question here is whether this should be 2.0.1 or rather 2.1.0 because we're adding new functionality?

Sorry for the elementary questions, but once I'm through this I should be of good help in the future in the maintenance of the package.

Jutho:
Another well polished PR. I'll try to review in more detail tomorrow. After having reviewed, the course of action is indeed to "Squash and merge", followed by drafting a new release (via the Github web interface), which I think can be `v2.1.0`, given the fact that substantial functionality has been added. `v2.0.1` would be for simple bug fixes. Attobot will automatically tag this new version in METADATA.jl. 

I seem to have messed something up with the settings of code coverage, since we don't get any update on coverage anymore.

dkarrasch:
Thanks, @Jutho. I'll advertise the new features in the README and add that to the PR. Regarding code coverage, I got the update by following up on the codecov-page, that's why the latest test commits. (I just double-checked, it seems to "edit" the above codecov post, the currently shown result is referring to the latest commit, so that's fine.) I'm not getting exactly how the codecov works, because some lines seem to be not covered which cannot be missed if the line above is called.

dkarrasch:
Haha, that's interesting. With the example in the README, for larger sizes (I tried 1000 and 3000) in `D`, neither of the `Arpack` methods converges for the default `maxiter`. The `IterativeSolvers.svdl` converges in both cases and is fast, I'd say. `TSVD.tsvd` converges only for size 1000x1000, but is slower by a factor of 4-5. Note, however, the difference in the error message:
```
ERROR: no convergence # in TSVD.jl
ERROR: ARPACKException: unspecified ARPACK error: 1 # Arpack.jl
```
Perhaps, there are different tolerance defaults behind the scenes, so this may not be fair. Anyway, I'm glad all these methods now fully collaborate with `LinearMaps.jl`.

Jutho:
The second derivative operator has a gap in the spectrum that scales as `1/N^2`, so iterative methods will have a hard time. Rather than increasing the number of iterations, I often find that increasing the Krylov dimension is more useful. I find the ARPACk default of krylov dimension 20 often too low for problems of medium size (i.e. vector length of a few thousands) but with a small gap.

Jutho:
My apologies for the delay, i hope to find some time for reviewing tomorrow.

dkarrasch:
No worries. Nothing depends yet on this PR, even though I hope that other packages will make good use of it some day. I don't think I have changed the code after your previous review, so it remains mostly to review the new README. If you know why code coverage doesn't recognize that those two or three lines are run and tested, that would be helpful, because I'm finding the codecov "failures" annoying. ;-) In my opinion, this package is 100% covered.

README.md
```diff
@@ -1,18 +1,22 @@
...
 
-*   Specifying the `eltype` of a function to be used as linear map should now use the constructor `LinearMap{T}(f, ...)`.
+## What's new in v2.1.0
+*   Fully Julia v0.7 compatible; dropped compatibility for previous versions of Julia from LinearMaps.jl v2.0.0 on.
+*   A 5-argument version for `mul!(y, A::LinearMap, x, α=1, β=0)`, which computes `y := α * A * x + β * y` and implements the usual 3-argument `mul!(y, A, x)` for the default `α` and `β`.
+*   Synonymous `convert(Matrix, A::LinearMap)` and `convert(Array, A::LinearMap)` functions, that call the `Matrix` constructor and return the matrix representation of `A`.
+*   Multiplication with matrices, interpreted as a block vector of vectors:
+    * `mul!(Y::AbstractArray, A::LinearMap, X::AbstractArray, α=1, β=0)`: applies `A` to each column of `X` and stores the result inplace in the corresponding column of `Y`;
+    * for the not-inplace multiplication, the approach is to compute `convert(Matrix, A * X)`; this is equivalent to applying `A` to each column of `X`. In generic code which handles both `A::AbstractMatrix` and `A::LinearMap`, the additional call to `convert` is a noop when `A` is a matrix.
```
Jutho: `not-inplace` -> `out-of-place`

Jutho:
One comment, otherwise looks good to merge

PR #29, opened 2018-10-22
Ensure all `Array`s are initialized with `undef`

cako:


dkarrasch:
Somehow my github seems to go crazy. Sometimes I do and sometimes I don't see your PR. Thanks for spotting that oversight.

dkarrasch:
Somehow my github seems to go crazy. Sometimes I do and sometimes I don't see your PR. Thanks for spotting that oversight.

Jutho:
I have the same problem, Github seems to do strange things. This is the first time I manage to view this PR. Thanks for merging.

Jutho:
And to @cako for committing of course.

dkarrasch:
First, I thought that the fixed methods are never called, because `transpose(::CompositeMap)` is a `CompositeMap` again. But it can get called in cases when `CompositeMap`s are first wrapped and then transposed. I have added tests for that in #30. Once that is merged (I can't access it currently because of the github issue :-D ), I'll tag a fix-release. Thanks again @cako.

Issue #24, opened 2018-07-10
Composition bug when chaining 4 LinearMaps


cako:
I ran into a bug when composing 4 maps using  LinearMaps 1.0.4. When running the following script in Julia 0.6.3, I get `ERROR: cannot resize array with shared data`.

```julia
import LinearMaps
dz, dx = 15, 15
X = 0:dx:1000; nx = length(X)
Z = 0:dz:1000; nz = length(Z)

ricker(to, f) = (1 - 2pi^2 * f^2 * to.^2) .* exp.(-pi^2 * f^2 * to.^2)
rick = ricker(Z-Z[div(nz-1,2)+1], 0.01); 

function conv_fwd(mod, wavelet, nz, nx)
    return hcat( (conv(reshape(mod, nz, nx)[:,ix], wavelet)[div(nz-1,2)+1:div(3nz-1,2)] for ix in 1:nx)... );
end
function conv_adj(mod, wavelet, nz, nx)
    return hcat( (conv(reshape(mod, nz, nx)[:,ix], wavelet[end:-1:1])[div(nz-1,2)+1:div(3nz-1,2)] for ix in 1:nx)... );
end

C = LinearMaps.LinearMap(x -> conv_fwd(x, rick, nz, nx)[:],
                         x -> conv_adj(x, rick, nz, nx)[:],
                         nz*nx, nz*nx)
x = ones(nz*nx)
(C*C*C*C*x)[10] == 0.888757975075548
```
The culprit is a `resize!` in `composition.jl` (line 93). I propose to simply create a new array at each composition multiplication (in a commit soon to follow). It is true this is not the most efficient solution, however, people using LinearMaps will tend to have very expensive operators such that the cost of allocation is negligible compared to the cost of multiplication.

dkarrasch:
Hi @cako , that was a tough one. Generally, this functionality is tested, so it came as a big surprise to me. The issue seems to be here with your function definition. As a MWE, take
```
src = ones(nz*nx)
dest = conv_fwd(src, rick, nz, nx)[:]
resize!(src, 20)
```
which yields
```
ERROR: cannot resize array with shared data
```
on my machine. So, applying your function to a `src` variable does something to it such that it cannot be resized afterwards. Regardless of whether that is meaningful, it is strange; after all, you would like to be able to do whatever you want to `src`. There are three options here:
[...]

PR #31, opened 2018-10-23
Add fallback when `resize!` does not work

cako:
I have been thinking about the issue #24 (using `resize!` within the composition operator) and have come to the conclusion that users should not be expected to always provide "clean" arrays which can be resized. Take my case for example, I had to essentially change my function by adding a couple of `copy` commands to ensure `resize!` was working. I was only able to do this after guidance. In fact, I have a more complex example (which I won't share because it is pretty long and specific) that I simply cannot get to work with the current code. Copying data both at the input and at the output do not work. I have [another (somewhat contrived) example](https://gist.github.com/cako/6b45d34eda9b4498ed8014a7659eaedb) demonstrates how this can happen. It can also be used for code coverage if the commit is accepted.

So either we can accept that some people's code will mess with the array in such a way that it cannot be resized, or we make them accept the fact that the library is not for them. I don't think the latter is beneficial for anyone!

On the other hand, I fully appreciate that `resize!` is much faster than array creation. I [benchmarked](https://gist.github.com/cako/2c42402d96cedbcf9b848704d79adc3a) resizing an array of size `1000` to several different sizes, ranging from `100` to `100_000`. The cost of only the resizing is in orange, while the cost of allocating from scratch is shown in green.
benchmark

Clearly allocating from scratch grows linearly, while resizing may eventually be linear as well, but it is offset by how much has already been allocated.

So I propose a very simple solution that I think will make everyone happy: we resize when we can, and when we can't, we allocate. I have added a simple try/catch block to do this. As shown in the figure above, this "fix" (blue line) retains decent performance for larger arrays, and is still much better than allocating for smaller arrays. It also has the obvious benefit of not breaking anyone's code!

The commits are very simple, simply wrap all `resize!` calls with a try/catch block which in the catch statement allocates an array.

### DIFF
--- a/src/composition.jl
+++ b/src/composition.jl
@@ -93,7 +93,11 @@ function A_mul_B!(y::AbstractVector, A::CompositeMap, x::AbstractVector)
             try
                 resize!(dest, size(A.maps[n], 1))
             catch err
-                dest = Array{T}(undef, size(A.maps[n], 1))
+                if err == ErrorException("cannot resize array with shared data")
+                    dest = Array{T}(undef, size(A.maps[n], 1))
+                else
+                    rethrow(err)
+                end
             end
             A_mul_B!(dest, A.maps[n], source)
             dest, source = source, dest # alternate dest and source
@@ -119,7 +123,11 @@ function At_mul_B!(y::AbstractVector, A::CompositeMap, x::AbstractVector)
             try
                 resize!(dest, size(A.maps[n], 2))
             catch err
-                dest = Array{T}(undef, size(A.maps[n], 2))
+                if err == ErrorException("cannot resize array with shared data")
+                    dest = Array{T}(undef, size(A.maps[n], 2))
+                else
+                    rethrow(err)
+                end
             end
             At_mul_B!(dest, A.maps[n], source)
             dest, source = source, dest # alternate dest and source
@@ -145,7 +153,11 @@ function Ac_mul_B!(y::AbstractVector, A::CompositeMap, x::AbstractVector)
             try
                 resize!(dest, size(A.maps[n], 2))
             catch err
-                dest = Array{T}(undef, size(A.maps[n], 2))
+                if err == ErrorException("cannot resize array with shared data")
+                    dest = Array{T}(undef, size(A.maps[n], 2))
+                else
+                    rethrow(err)
+                end
             end
             Ac_mul_B!(dest, A.maps[n], source)
             dest, source = source, dest # alternate dest and source
--- a/src/composition.jl
+++ b/src/composition.jl
@@ -90,7 +90,11 @@ function A_mul_B!(y::AbstractVector, A::CompositeMap, x::AbstractVector)
             dest = Array{T}(undef, size(A.maps[2], 1))
         end
         for n=2:N-1
-            resize!(dest, size(A.maps[n], 1))
+            try
+                resize!(dest, size(A.maps[n], 1))
+            catch err
+                dest = Array{T}(undef, size(A.maps[n], 1))
+            end
             A_mul_B!(dest, A.maps[n], source)
             dest, source = source, dest # alternate dest and source
         end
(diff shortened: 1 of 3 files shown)


Jutho:
`reshape` is indeed a very common use case to trigger this, and I agree this is a bug.

I think something like this can detect whether an `Array` is shared, and thus, if `resize!` will fail
```julia
_flags(A::Array) = unsafe_load(convert(Ptr{UInt16}, pointer_from_objref(A)), 9)
_isshared(A::Array) = !(_flags(A) & 0x4000 == 0x0000)
```
I am however a bit worried if this depends on the architecture (i.e. 32 vs 64 bits), in particular the value 9 in the first function will probably have to be different on 32 bit. So probably it needs to be
```julia
if Sys.WORD_SIZE == 64
 _flags(A::Array) = unsafe_load(convert(Ptr{UInt16}, pointer_from_objref(A)), 9)
elseif Sys.WORD_SIZE == 32
 _flags(A::Array) = unsafe_load(convert(Ptr{UInt16}, pointer_from_objref(A)), 5)
else
 error("Unknown word size")
end
_isshared(A::Array) = !(_flags(A) & 0x4000 == 0x0000)
```
But I am not sure and don't have a 32 bit machine to test this.

dkarrasch:
That sounds like a plan: detecting the specific `reshape` problem and adjusting the code for this case, instead of a generic try-catch-block. Let's what the experts say in the `push!/pop!` issue...

Jutho:
Another way out is that we pass `view(dest, :)` to the user's linear map. This can never lead to shared data, but upon reshaping it creates a `ReshapedArray`. Depending on what the linear map does, this might also have a small performance cost. Furthermore, this assumes that the user did write a function that accepts `AbstractVector` arguments instead of just `Vector`.

dkarrasch:
Regardless of how we're going to fix this issue, I think adding the test would be good. This is a generic problem that we would want to be able to solve.

Unfortunately, there is not much to be learned from the implementation of `resize!`, the error is thrown from the inside of some `ccall`. It is not thrown when you have arrays with shared data that happen to keep the size. For these cases we need to make sure not to add significant overload. Is that what your simulations are showing, @cako? Are you taking some good old "standard" case and measure runtime with and without try-catch etc.? Could you perhaps share your benchmarking code in the gist?

Jutho:
> This is where I start getting confused, in 32 bits, void* and size_t should have size 4 bytes each. In 64 bits, they should have 8 bytes each, placing the memory location of jl_array_flags_t for the respective architectures at 9 and 17. However, you have indices 5 and 9 for that memory location.

@cako, the numbers 5 and 9 are measured in units of `sizeof(UInt16)`, because the pointer is a `Ptr{UInt16}`. I convert the pointer to a `Ptr{UInt16}` because I know from the source code of `array.c` that the flags field in the array struct is 16 bits wide, with the isshared flag being on bit 15 out of 16, see here: https://github.com/JuliaLang/julia/blob/master/src/julia.h#L150

But I agree that this approach is too fragile and clearly advised against.

The `try .... catch` solution is certainly fine with me, I think the only other solution is giving the user `view`s into the array. Or a `unsafe_wrap(Array, pointer(dest), ...)`, but that's again becoming "unsafe"!

dkarrasch:
Yes, I think that's a good compromise. In my own (unrelated to this package) code, I used try-catch-blocks (perhaps in a bad way), and had difficulties of finding mistakes after changes in the code, and just got `Nothing`. So that's my own personal bad experience with it. That's why I think it is good to be as specific as possible about the error that one wants to catch (as you suggested), in order to not cover up other issues which the user should consider and solve.

dkarrasch:
Thanks @cako for the nice tests. I made some minor changes. The transpose/adjoint of a CompositeMap is a CompositeMap again, and therefore gets handled by the `A_mul_B!`, and not the `A*_mul_B!`s. I hope this makes coverage look better.

dkarrasch:
@Jutho: While @cako was working on his branch, our master branch diverged slightly (in runtests only). How do I make sure that our latest tests and @cako's new tests both survive the merge? Is that what rebasing is about? With the union of the tests, we should be able to keep excellent coverage.

Jutho:
Since Github still states: "This branch has no conflicts with the base branch" I think this can still safely be merged in master and all edits (commits) will survive. But we can also try to rebase, this shows up even on Github, if you fold out the "Squash and merge" drop down menu.

dkarrasch:
I wasn't sure what is meant by "base branch", if it's our current master or the one where @cako branched off. I'll simply squash, merge, and tag a new fix release 2.2.2 then.

Jutho:
Alternatively, @cako could rebase locally on the latest master and then push (maybe push -f for forced) here

dkarrasch:
No worries. It's a bit tricky (and in fact, I was believing up until recently that those methods are not needed at all ;-) ).

The rebase is mainly to have all tests together, let code coverage run again and get good grades.

PR #34, opened 2018-10-29
avoid allocation for length-1 linear combinations

dkarrasch:


Jutho:
Looks good; if Julia would expose the true BLAS functionality of `gemv`, which can do ``y = beta*y+ alpha* A * x`` , we wouldn't need a temporary at all.

dkarrasch:
I think we need it for `FunctionMap`s anyway, don't we? We can't compute the image vector "componentwise", so we need to store the `A*x` somewhere. But now that you say it, I think there is a 5-args `mul!` somewhere in Base, at least there used to be. Certainly, we have it for LinearMaps. I'll rewrite it and see if it works with `WrappedMap`s. I can't seem to find a 5-args method for `mul!` in LinearAlgebra, and I remember an issue in `IterativeSolvers.jl`.

Jutho:
Indeed for `FunctionMap`s we do, unless we require the user to implement a 5 argument `mul!`. I was a bit too fast here :)

Jutho:
This looks good. Is this ready to be merged?

chriscoey:
Discussion on the 5-arg mul! seems to have stalled, which makes me sad

dkarrasch:
Yes, I think this can be merged. We were seriously hit by code coverage changes, so we dropped down to 60%. I started a major revision of the tests, which has significantly improved coverage. I wanted to take another look at it, but I can do it in a separate PR (I have some other minor things to suggest).

dkarrasch:
Regarding the 5-arg mul!, we will need to be very careful. Our current `mul!` for `LinearCombination`s allocates exactly once (`z = similar(y)`) when at least two `LinearMap`s are combined (given that the `LinearMap` is built from matrices and/or mutating `FunctionMap`s). If we replace our current "3-arg mul" `A_mul_B!`-`axpy!`-combination by a 5-arg mul! in each step, we may get rid of any extra allocation in case everything is built from matrices (and the 5-arg mul! is provided, of course), but we would allocate once per linear map in the linear combination, see our own current 5-arg mul! which allocates `A*x` at some point. So, either we find a nice workaround, or live with our current version, which would be not too bad.

PR #35, opened 2018-10-30
Some minor simplifications

dkarrasch:
* generic issymmetric, ishermitian, and isposdef is never used, so removed

* AdjointMap's cannot be generated from LinearMap{<:Real}, comparison yields always false

* added more tests to increase coverage

### DIFF
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -10,6 +10,10 @@ abstract type LinearMap{T} end
 Base.eltype(::LinearMap{T}) where {T} = T
 
 Base.isreal(A::LinearMap) = eltype(A) <: Real
+LinearAlgebra.issymmetric(::LinearMap) = false # default assumptions
+LinearAlgebra.ishermitian(A::LinearMap{<:Real}) = issymmetric(A)
+LinearAlgebra.ishermitian(::LinearMap) = false # default assumptions
+LinearAlgebra.isposdef(::LinearMap) = false # default assumptions
 
 Base.ndims(::LinearMap) = 2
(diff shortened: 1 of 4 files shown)


dkarrasch:
@Jutho: I'm tired of working on the tests, so I'd leave it as is, and merge if you're ok with the other changes. We don't need a version tag, since this doesn't alter any functionality.

src/LinearMaps.jl
```diff
@@ -10,10 +10,6 @@ abstract type LinearMap{T} end
 Base.eltype(::LinearMap{T}) where {T} = T
 
 Base.isreal(A::LinearMap) = eltype(A) <: Real
-LinearAlgebra.issymmetric(::LinearMap) = false # default assumptions
-LinearAlgebra.ishermitian(A::LinearMap{<:Real}) = issymmetric(A)
-LinearAlgebra.ishermitian(::LinearMap) = false # default assumptions
-LinearAlgebra.isposdef(::LinearMap) = false # default assumptions
```
Jutho: I think these fallbacks were for people that write their own `LinearMap` subtype, and don't define these properties. A test case could include a `SimpleFunctionMap` type that does not declare these properties.

Jutho:
I also don't like writing tests very much, so thanks for all the work you've already done. This package has received much more attention since you became a collaborator.

src/LinearMaps.jl
dkarrasch: I see, I didn't think that far. :) I'll add them back again... and add some tests. 😆

dkarrasch:
Testing is like sports, higher grades, higher percentage of coverage. Initially you're enthusiastic, when you hit one code line by one test. But then it gets frustrating, when you simply don't know how to hit the uncovered lines. 🤣 

Anyway, now the generic structure functions are back and tested. I'll merge now.

PR #37, opened 2019-01-01
WIP/RFC: add support for noncommutative number types

### LINKED ISSUE #33: Convert coefficients in linear combinations to promote_type?
I am working on a minor improvement in our handling of linear combinations, i.e., avoiding allocation of `z = similar(y)` when the length of `A.maps` is 1. Such linear combinations occur when we multiply a linear map with a scalar factor, for instance. This is how I realized that we don't convert the coefficients to the `promote_type`, take e.g.
```
function Base.:(*)(α::Number, A::LinearMap)
    T = promote_type(eltype(α), eltype(A))
    return LinearCombination{T}(tuple(A), tuple(α))
end
```
Or rather, we leave it to the `lmul!` in `A*_mul_B!` to do this, well, at every multiplication, which may be many times in an iterative algorithm. As a quick benchmark, consider
```
using BenchmarkTools, LinearAlgebra
@btime lmul!($(rand()),$(rand(Complex{Float64}, 5000))) #  199.095 μs (0 allocations: 0 bytes)
@btime lmul!($(rand(Complex{Float64})),$(rand(Complex{Float64}, 5000))) # 188.849 μs (0 allocations: 0 bytes)
```
Confusingly, the opposite effect occurs here:
```
α = 2
@btime lmul!($(convert(Int64, α)), $(rand(5000))) #  523.188 ns (0 allocations: 0 bytes)
@btime lmul!($(convert(Float64, α)), $(rand(5000))) # 868.077 ns (0 allocations: 0 bytes)
```
Admittedly, it is exactly `lmul!` that seems to be affected by recent changes in `LinearAlgebra`, so this may need testing on nightly? I'm on standard
```
julia> versioninfo()
Julia Version 1.0.1
Commit 0d713926f8 (2018-09-29 19:05 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, skylake)
Environment:
  JULIA_EDITOR = atom -a
  JULIA_NUM_THREADS = 4
```
Any opinions?

### LINKED ISSUE #36: LinearMaps with UniformScaling
Is there any reason that a linear map shouldn't work with a UniformScaling from the LinearAlgebra package? Something like:
```
Base.:(*)(A::LinearMap, α::UniformScaling) = *(α.λ, A)
Base.:(*)(α::UniformScaling, A::LinearMap) = *(α.λ, A)
Base.:(*)(A::LinearMaps.LinearCombination, α::UniformScaling) = *(α.λ, A)
Base.:(*)(α::UniformScaling, A::LinearMaps.LinearCombination) = *(α.λ, A)
```
I find this useful, but perhaps there is a reason it is not already done?

dkarrasch:
Fixes #36, closes #33.

The `IdentityMap` type has become obsolete, it was used for addition of `LinearMap`s and `UniformScaling`s only. This functionality is superseded by the `UniformScalingMap` type.

TODOs:

- [x] tests for some noncommutative number type (include some additional dependency for testing? copy some minimal quaternion code from Julia base?)
- [x] delete identitymap.jl
- [x] reduce subsequent scalar multiplications


Jutho:
I have not been following this; will you let me know when this is ready for a review?

dkarrasch:
In some sense, this is ready for review. I have local tests using `Quaternions.jl`. I'm not sure if that's a problem with having a new dependency in the tests, but I wasn't able to get the short `testhelpers/Quaternions.jl` running with `LinearMaps`. So if you don't see any issue with the dependency, I guess all I need to do is to put a `REQUIRE` file in the tests folder, that would list LinearMaps and Quaternions.

Another very subtle issue is the following. In Julia.Base, there is a horrible typo in `mul!(Y, s::Number, X)`, because it calls `generic_mul!(Y, X, s)` which multiplies from the right, whereas we intended to multiply from the left. @tkf identified this typo and is going to fix it in the 5-arg-`mul!` PR once that is merged. The question is what to do here? At some point, there will be a fix in Julia (nightly -> latest -> 1.2 or whatever), but for other versions there won't. Does that mean we have to mess around with version numbers, or Compat? In my local package (to be commited to this PR), I have reversed the order in `mul!(Y, ::UniformScalingMap, X)`, which looks as confusing as the bug in Julia.Base. Another option would be to call `LinearAlgebra.generic_mul!` directly, which uses the intended order of multiplication, but I guess it's not so nice (even though I have to admit that I don't fully understand that distinction in `mul!` and `generic_mul!`; the signatures seem to be exactly the same).

Anyway, I'll push the noncommutative tests along with a test-REQUIRE and the "fix" of the order of multiplication. Tests pass for this combination, and then we can discuss how to continue.

Jutho:
> In some sense, this is ready for review. I have local tests using Quaternions.jl. I'm not sure if that's a problem with having a new dependency in the tests, but I wasn't able to get the short testhelpers/Quaternions.jl running with LinearMaps. So if you don't see any issue with the dependency, I guess all I need to do is to put a REQUIRE file in the tests folder, that would list LinearMaps and Quaternions.

I have no problem with another requirement in the tests only.

> Another very subtle issue is the following. In Julia.Base, there is a horrible typo in mul!(Y, s::Number, X), because it calls generic_mul!(Y, X, s) which multiplies from the right, whereas we intended to multiply from the left. @tkf identified this typo and is going to fix it in the 5-arg-mul! PR once that is merged. The question is what to do here? At some point, there will be a fix in Julia (nightly -> latest -> 1.2 or whatever), but for other versions there won't. Does that mean we have to mess around with version numbers, or Compat? In my local package (to be commited to this PR), I have reversed the order in mul!(Y, ::UniformScalingMap, X), which looks as confusing as the bug in Julia.Base. Another option would be to call LinearAlgebra.generic_mul! directly, which uses the intended order of multiplication, but I guess it's not so nice (even though I have to admit that I don't fully understand that distinction in mul! and generic_mul!; the signatures seem to be exactly the same).

I would not be opposed to ignoring the bug in Julia Base and doing the correct thing in this package.

src/linearcombination.jl
```diff
@@ -1,136 +1,86 @@
...
     # no size checking, will be done by individual maps
     A_mul_B!(y, A.maps[1], x)
-    A.coeffs[1] == 1 || lmul!(A.coeffs[1], y)
     l = length(A.maps)
     if l>1
         z = similar(y)
         for n=2:l
             A_mul_B!(z, A.maps[n], x)
-            axpy!(A.coeffs[n], z, y)
+            y .+= z
```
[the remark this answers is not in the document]
dkarrasch: Technically, we have our own 5-arg mul!, which would seem to do it in one pass. But for general `LinearMap`s, this would allocate a `z` in each call internally, and for matrices this is only in the making. I assume that allocating everytime is more expensive than visiting the preallocated components twice. But I agree, we should keep the idea of having an efficient "muladd!" in mind.

src/composition.jl
```diff
@@ -18,35 +18,42 @@ Base.size(A::CompositeMap) = (size(A.maps[end], 1), size(A.maps[1], 2))
...
-function LinearAlgebra.ishermitian(A::CompositeMap)
-    N = length(A.maps)
-    if isodd(N)
-        ishermitian(A.maps[div(N+1, 2)]) || return false
-    end
-    for n = 1:div(N, 2)
-        A.maps[n] == adjoint(A.maps[N-n+1]) || return false
+LinearAlgebra.ishermitian(A::CompositeMap) = _ishermitian(A.maps)
+_ishermitian(maps::Tuple{}) = true
+_ishermitian(maps::Tuple{<:LinearMap}) = issymmetric(maps[1])
```
dkarrasch: Do you really mean `issymmetric`here, or is this an oversight from copy-paste-modify?

Jutho: Very much a typo from copy paste. I've pushed a new version that is more robust against copy paste errors.

dkarrasch:
Once I have your update, I'll push my version that uses `LinearAlgebra.generic_mul!` in `UniformScalingMap`-mul! with arguments in the expected order. This will work on all Julia versions. At some point, we may wish to change that back to `mul!` and lower bound the Julia version for that. Also, I've added inference tests to almost all tests to enforce type stability in future development. With that, I guess we can squash, merge and tag (a minor release)?

Jutho:
Absolutely, go ahead :-).

dkarrasch:
I noticed one curious thing. For a random quaternion matrix `A`, `A'A` is numerically determined as not hermitian! The reason is that `(A'A)[1,2] == conj((A'A)[2,1])` fails, but `(A'A)' ≈ (A'A)` passes. So, `LinearMaps.jl` is smarter than Julia! 😆
```
using Quaternions
A = Quaternion.(rand(10,10), rand(10,10), rand(10,10), rand(10,10))
L = LinearMap(A)
ishermitian(L'L) # true
ishermitian(A'A) # false
(A'A)' ≈ A'A # true
```

PR #39, opened 2019-01-19
Create getindex.jl to support A[:,j] etc.

JeffFessler:


JeffFessler:
sorry about the coverage issue - i wrote tests that cover all of the new functions, but i put them in a separate file in a separate pull request...

PR #41, opened 2019-02-23
Testings on Octonions and more testings on quaternions

yikait2:


dkarrasch:
As I said in #41, testing `Octonion`s is okay, but one would not expect composite maps to work as written in the tests. Maybe we should leave it at `Quaternion`s then? Do you think we should have more tests for them?

dkarrasch:
Bump. I couldn't find new tests for the Quaternion case. Do you have some more? Otherwise I'd close the issue, and maybe add a note in the README that this package only works properly for associative element types.

dkarrasch:
The test failure on nightly is due to the missing `float(::Octonion)` function in `Quaternions.jl`. This is going to be fixed upstream.

PR #47, opened 2019-06-06
update to new registration process

dkarrasch:
Generating the `Project.toml` turned out to be straightforward. I took the freedom to set the lower bound to v1.0 for Julia. @Jutho, you may wish to add an author/e-mail field or whatever, and then we'll probably need to set a version field. I suppose this constitutes a minor update, so it should be `version = "2.4.0"`?

For reference, let me post this link to the registrator page.

https://github.com/JuliaRegistries/Registrator.jl

IIUC, we need to install that app (as a replacement for attobot). Then it's up to us if we also want to use TagBot, or whether we do the tags manually (which IMHO should be fine for this package).


Jutho:
Registrator is already installed and should have access to LinearMaps.jl. I am not sure an author field is necessary, but I could add one. 

There also seem to be some test changes/additions in this PR? (which is certainly a good thing)

dkarrasch:
I didn't feel like not including an author field without even asking. So I'll leave it up to you.

Yes, I had some tests lying around which I presumably wrote shortly after the "noncommutative" change, or in face of the Octonion discussion (I don't remember), that I hadn't committed earlier. I replaced the home-made allocation-counting macro by the probably better maintained BenchmarkTools solution.

Since we're not changing any functionality, do we have to lift up the version number at all? Can't we register with the current version? Or don't we need to register because they copied existing METADATA entries to the GeneralRegistry? If in doubt, we could probably simply "release" v2.4.0 and register via the Registrator, to enter the new process.

Jutho:
Certainly the version restriction from julia 0.7 to now >= 1 implies that we bump the version number, and even that it be more than a patch update. A minor update is fine though, I believe.

chriscoey:
bump :)

dkarrasch:
@JuliaRegistrator register()

PR #48, opened 2019-07-02
disallow test failures on Julia v1.2

dkarrasch:
The missing `float` method for `Octonion` has been merged and tagged in `Quaternions.jl`, so there is no reason to allow test failure on Julia v1.2.


Jutho:
Looks good.

PR #49, opened 2019-07-10
UniformScalingMap improvements

dkarrasch:
* speed up mul with UniformScalingMap for lambda=0, 1
* inherit type from LinearMap

### DIFF
--- a/src/uniformscalingmap.jl
+++ b/src/uniformscalingmap.jl
@@ -19,7 +19,9 @@ LinearAlgebra.transpose(A::UniformScalingMap) = A
 LinearAlgebra.adjoint(A::UniformScalingMap)   = UniformScalingMap(conj(A.λ), size(A))
 
 # multiplication with vector
-Base.:(*)(A::UniformScalingMap, x::AbstractVector) = A.λ * x
+Base.:(*)(A::UniformScalingMap, x::AbstractVector) =
+    length(x) == A.M ? A.λ * x : throw(DimensionMismatch("A_mul_B!"))
+
 # call of LinearAlgebra.generic_mul! since order of arguments in mul! in stdlib/LinearAlgebra/src/generic.jl
 # TODO: either leave it as is or use mul! (and lower bound on version) once fixed in LinearAlgebra
 function A_mul_B!(y::AbstractVector, A::UniformScalingMap, x::AbstractVector)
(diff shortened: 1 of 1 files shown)


dkarrasch:
@Jutho If you find the time, can you please take a look at this. The interesting part here is at the end, where the type `T` is inferred. I may have misunderstood something, but I think `T` was in the wrong place, also back in the `IdentityMap` times. So we should make sure this all correct. Then also the mul-improvements should speed up some trivial cases, which may become somewhat relevant in the block maps case, when people have plain identity blocks.

dkarrasch:
I think I was mislead by the comment in the very first line. It says
```julia
# T will be determined from maps to which this is added
```
which is inconsistent with what we currently have, because `T` is determined from the `UniformScaling`, not from the `LinearMap`. However, the resulting `LinearCombination` `L+I` gets the promoted type. But I think it is of no harm to determine `T` from the `UniformScaling`. So we may remove those changes (and the comment), or leave the comment and keep the changes that I have. After all, the `convert` affects a single number only.

src/uniformscalingmap.jl
```diff
@@ -19,20 +19,48 @@ LinearAlgebra.transpose(A::UniformScalingMap) = A
...
+function At_mul_B!(y::AbstractVector, A::UniformScalingMap, x::AbstractVector)
+    (length(x) == length(y) == A.M || throw(DimensionMismatch("At_mul_B!")))
+    if iszero(A.λ)
+        return fill!(y, 0)
+    elseif isone(A.λ)
+        return copyto!(y, x)
+    else
+        return LinearAlgebra.generic_mul!(y, A.λ, x)
+    end
+end
```
Jutho: Maybe we can just have
```julia
At_mul_B!(y::AbstractVector, A::UniformScalingMap, x::AbstractVector) = A_mul_B!(y, transpose(A), x)
Ac_mul_B!(y::AbstractVector, A::UniformScalingMap, x::AbstractVector) = A_mul_B!(y, adjoint(A), x)
```

src/uniformscalingmap.jl
```diff
@@ -19,20 +19,48 @@ LinearAlgebra.transpose(A::UniformScalingMap) = A
...
 
 # combine LinearMap and UniformScaling objects in linear combinations
-Base.:(+)(A1::LinearMap, A2::UniformScaling{T}) where {T} = A1 + UniformScalingMap{T}(A2[1,1], size(A1, 1))
-Base.:(+)(A1::UniformScaling{T}, A2::LinearMap) where {T} = UniformScalingMap{T}(A1[1,1], size(A2, 1)) + A2
-Base.:(-)(A1::LinearMap, A2::UniformScaling{T}) where {T} = A1 - UniformScalingMap{T}(A2[1,1], size(A1, 1))
-Base.:(-)(A1::UniformScaling{T}, A2::LinearMap) where {T} = UniformScalingMap{T}(A1[1,1], size(A2, 1)) - A2
+Base.:(+)(A1::LinearMap{T}, A2::UniformScaling) where {T} =
+    A1 + UniformScalingMap(convert(T, A2[1,1]), size(A1, 1))
```
Jutho: This looks wrong; there is no guarantee that `A2[1,1]` (i.e. lambda) can be converted to type `T`. Some promotion should be involved at least.

src/uniformscalingmap.jl
```diff
@@ -19,20 +19,48 @@ LinearAlgebra.transpose(A::UniformScalingMap) = A
 LinearAlgebra.adjoint(A::UniformScalingMap)   = UniformScalingMap(conj(A.λ), size(A))
 
 # multiplication with vector
+Base.:(*)(A::UniformScalingMap, x::AbstractVector) = A.λ * x
```
Jutho: Unlike the mutating version, this does not check whether the vector length is compatible (not your fault, was so in the original version). But maybe such a check can be added for consistency?

src/uniformscalingmap.jl
dkarrasch: Yes, since `transpose` and `adjoint` don't add another wrapping layer for `UniformScalingMap`s, this should work.

src/uniformscalingmap.jl
dkarrasch: I'm not sure I fully understand the difference between conversion and promotion. The first sentence in the Julia docs for promotion reads

> Promotion refers to converting values of mixed types to a single common type.

To me, this sounds like first the types of the arguments of `promote` are extracted, then the types are promoted, and eventually all arguments are converted to the common type. Do you have an example of what would not work with `convert`, but with `promote`?

src/uniformscalingmap.jl
dkarrasch: Aha, seems like this should also work for other "invariant" types under adjoint/transpose like `CompositeMaps` and `LinearCombination`, and eventually `BlockMaps`! 🎉

src/uniformscalingmap.jl
Jutho: yes, if `T<:Real` and the `UniformScaling` parameter is complex, you will get an error.

If `T == Float32` and the `UniformScaling` is `Float64`, promotion would result in both being `FLoat64`, whereas the current approach will convert to `Float32`.

dkarrasch: Ok, thanks, now I see. So we do something like
```julia
convert(promote_type(T, eltype(A2)), A2[1,1])
```
? And maybe I'll replace the `A2[1,1]` by `A2.λ`.

Jutho: Sure, but the old version was also working fine, or not? I am fine either way, will amount to more or less the same thing.

dkarrasch: Yes it was, but the lambda is more direct... but probably this would get compiled away anyway. 😄

src/uniformscalingmap.jl
```diff
@@ -19,20 +19,30 @@ LinearAlgebra.transpose(A::UniformScalingMap) = A
...
+At_mul_B!(y::AbstractVector, A::UniformScalingMap, x::AbstractVector) = A_mul_B!(y, transpose(A), x)
+Ac_mul_B!(y::AbstractVector, A::UniformScalingMap, x::AbstractVector) = A_mul_B!(y, adjoint(A), x)
 
 # combine LinearMap and UniformScaling objects in linear combinations
-Base.:(+)(A1::LinearMap, A2::UniformScaling{T}) where {T} = A1 + UniformScalingMap{T}(A2[1,1], size(A1, 1))
-Base.:(+)(A1::UniformScaling{T}, A2::LinearMap) where {T} = UniformScalingMap{T}(A1[1,1], size(A2, 1)) + A2
-Base.:(-)(A1::LinearMap, A2::UniformScaling{T}) where {T} = A1 - UniformScalingMap{T}(A2[1,1], size(A1, 1))
-Base.:(-)(A1::UniformScaling{T}, A2::LinearMap) where {T} = UniformScalingMap{T}(A1[1,1], size(A2, 1)) - A2
+Base.:(+)(A1::LinearMap{T}, A2::UniformScaling) where {T} =
+    A1 + UniformScalingMap(convert(promote_type(T, eltype(A2)), A2.λ), size(A1, 1))
```
Jutho: Looks good, but still a bit asymmetric though. It could just as well be:
```julia
Base.:(+)(A1::LinearMap, A2::UniformScaling =
	    A1 + UniformScalingMap(convert(promote_type(eltype(A1), eltype(A2)), A2.λ), size(A1, 1))
```

Jutho:
Looks good; I'll merge.

PR #51, opened 2019-07-19
Simplify multiplication code

dkarrasch:
This PR reduces multiplication code for types that are "invariant" under adjoint/transpose, analogous to what @Jutho suggested in #49 for `UniformScalingMap`s.

In the subspace-mul, I use Julia Base's `eachcol` function, which creates an iterator over column views.


dkarrasch:
I overlooked that `eachcol` is a new feature in Julia v1.1, so I removed the use of it, but left a comment in case we're ever going to consider supporting Julia versions above 1.1 exclusively.

Jutho:
Looks great; strange that I never earlier had the idea to do it like this :-).

PR #52, opened 2019-07-27
Optimize block map construction and multiplication

dkarrasch:
I couldn't stop myself from further optimizing things, partly for my own education (@JeffFessler 😉 ). Let me start with a simple "benchmark problem".
```julia
using LinearMaps, LinearAlgebra, BenchmarkTools
A = LinearMap(rand(10, 10))
L = hvcat((2, 2), A, A, A, I)
x = rand(size(L, 2))
y = rand(size(L, 1))
@btime mul!($y, $L, $x)
M = Matrix(L)
@btime mul!($y, $M, $x)
```
On current master, we get:
```julia
 2.682 μs (33 allocations: 2.23 KiB)
  97.825 ns (0 allocations: 0 bytes)
```
On this branch, we get:
```julia
 1.365 μs (17 allocations: 896 bytes)
  96.926 ns (0 allocations: 0 bytes)
```
So, a factor of 2 on this small problem. What has happened? I changed from the `Vector`-based `cumsum` formulation of obtaining the starting indices corresponding to blocks, and changed it to a tuple-based, recursive, inferable formulation. Obviously, the values obtained are the exact same in all cases. This is responsible for the major improvement above. This is compiler magic that I haven't expected myself.

What else? We are allocating memory because we generally can't compute the target vectors component-wise. So when we add the result of some block multiplication to a previous result, we need to first allocate the intermediate result before we can add it. This is, in general, unavoidable. BUT, there are two exceptions: uniform scalings (for which I have an allocation-free (this is tested!), 5-arg `mul!`), and wrapped maps with BLAS-matrices. Unfortunately, we don't have that 5-arg mul! for (dense) matrices, yet. But the minute that is coming, our code will perform better immediately. The remaining allocations are likely to generate views. For instance
```julia
L = [A I I]
x = rand(size(L, 2))
y = rand(size(L, 1))
@btime mul!($y, $L, $x)
```
should not be allocating at all, because the identities are handled by the allocation-free 5-arg `mul!`, but we still get
```julia
 687.351 ns (14 allocations: 624 bytes)
```
in comparison to
```julia
M = Matrix(L)
@btime mul!($y, $M, $x)
133.692 ns (0 allocations: 0 bytes)
```
In good cases, I think the overhead is pretty much constant, and is likely to be irrelevant when costly matrix multiplication is involved.

Other minor improvements include that size checks are now done at linear map promotion.

### DIFF
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -11,13 +11,8 @@ end
 
 BlockMap{T}(maps::As, rows::S) where {T,As<:Tuple{Vararg{LinearMap}},S} = BlockMap{T,As,S}(maps, rows)
 
-# firstindices(maps::Tuple{Vararg{LinearMap}}, dim) = cumsum([1, map(m -> size(m, dim), maps)...,])
-
-function check_dims(maps::Tuple{Vararg{LinearMap}}, k)
-    n = size(maps[1], k)
-    for map in maps
-        n == size(map, k) || throw(DimensionMismatch("Expected $n, got $(size(map, k))"))
-    end
+function check_dim(A::LinearMap, dim, n)
+    n == size(A, dim) || throw(DimensionMismatch("Expected $n, got $(size(A, dim))"))
     return nothing
 end
 
@@ -73,9 +68,7 @@ function Base.hcat(As::Union{LinearMap,UniformScaling}...)
     end
     nrows == 0 && throw(ArgumentError("hcat of only UniformScaling-like objects cannot determine the linear map size"))
 
-    maps = promote_to_lmaps(ntuple(i->nrows, nbc), 1, As...)
-    check_dims(maps, 1)
-    return BlockMap{T}(maps, (length(As),))
+    return BlockMap{T}(promote_to_lmaps(ntuple(i->nrows, nbc), 1, 1, As...), (nbc,))
 end
 
 ############
@@ -96,9 +89,7 @@ function Base.vcat(As::Union{LinearMap,UniformScaling}...)
     end
     ncols == 0 && throw(ArgumentError("hcat of only UniformScaling-like objects cannot determine the linear map size"))
 
-    maps = promote_to_lmaps(ntuple(i->ncols, nbr), 1, As...)
-    check_dims(maps, 2)
-    return BlockMap{T}(maps, ntuple(i->1, length(As)))
+    return BlockMap{T}(promote_to_lmaps(ntuple(i->ncols, nbr), 1, 2, As...), ntuple(i->1, nbr))
 end
 
 ############
@@ -157,18 +148,18 @@ function Base.hvcat(rows::NTuple{nr,Int}, As::Union{LinearMap,UniformScaling}...
         end
     end
 
-    return BlockMap{T}(promote_to_lmaps(n, 1, As...), rows)
+    return BlockMap{T}(promote_to_lmaps(n, 1, 1, As...), rows)
 end
 
-promote_to_lmaps_(n::Int, J::UniformScaling) = UniformScalingMap(J.λ, n)
-promote_to_lmaps_(n::Int, A::LinearMap) = A
+promote_to_lmaps_(n::Int, dim, J::UniformScaling) = UniformScalingMap(J.λ, n)
+promote_to_lmaps_(n::Int, dim, A::LinearMap) = (check_dim(A, dim, n); A)
 promote_to_lmaps(n, k) = ()
-promote_to_lmaps(n, k, A) = (promote_to_lmaps_(n[k], A),)
-promote_to_lmaps(n, k, A, B) = (promote_to_lmaps_(n[k], A), promote_to_lmaps_(n[k+1], B))
-promote_to_lmaps(n, k, A, B, C) =
-    (promote_to_lmaps_(n[k], A), promote_to_lmaps_(n[k+1], B), promote_to_lmaps_(n[k+2], C))
-promote_to_lmaps(n, k, A, B, Cs...) =
-    (promote_to_lmaps_(n[k], A), promote_to_lmaps_(n[k+1], B), promote_to_lmaps(n, k+2, Cs...)...)
+promote_to_lmaps(n, k, dim, A) = (promote_to_lmaps_(n[k], dim, A),)
+promote_to_lmaps(n, k, dim, A, B) = (promote_to_lmaps_(n[k], dim, A), promote_to_lmaps_(n[k+1], dim, B))
+promote_to_lmaps(n, k, dim, A, B, C) =
+    (promote_to_lmaps_(n[k], dim, A), promote_to_lmaps_(n[k+1], dim, B), promote_to_lmaps_(n[k+2], dim, C))
+promote_to_lmaps(n, k, dim, A, B, Cs...) =
+    (promote_to_lmaps_(n[k], dim, A), promote_to_lmaps_(n[k+1], dim, B), promote_to_lmaps(n, k+2, dim, Cs...)...)
 
 ############
 # basic methods
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -215,7 +215,7 @@ function A_mul_B!(y::AbstractVector, A::BlockMap, x::AbstractVector)
     yinds = firstrowindices(A)
     @views for rowind in 1:length(rows)
         xinds = firstcolindices(maps[mapind+1:mapind+rows[rowind]])
-        yrow = @views y[yinds[rowind]:(yinds[rowind+1]-1)]
+        yrow = y[yinds[rowind]:(yinds[rowind+1]-1)]
         mapind += 1
         A_mul_B!(yrow, maps[mapind], x[xinds[1]:xinds[2]-1])
         for colind in 2:rows[rowind]
@@ -241,7 +241,7 @@ function At_mul_B!(y::AbstractVector, A::BlockMap, x::AbstractVector)
     # subsequent block rows, add results to corresponding parts of y
     @views for rowind in 2:length(rows)
         yinds = firstcolindices(maps[mapind+1:mapind+rows[rowind]])
-        xcol = @views x[xinds[rowind]:(xinds[rowind+1]-1)]
+        xcol = x[xinds[rowind]:(xinds[rowind+1]-1)]
         for colind in 1:rows[rowind]
             mapind +=1
             mul!(y[yinds[colind]:yinds[colind+1]-1], transpose(maps[mapind]), xcol, 1, 1)
@@ -265,7 +265,7 @@ function Ac_mul_B!(y::AbstractVector, A::BlockMap, x::AbstractVector)
     # subsequent block rows, add results to corresponding parts of y
     @views for rowind in 2:length(rows)
         yinds = firstcolindices(maps[mapind+1:mapind+rows[rowind]])
-        xcol = @views x[xinds[rowind]:(xinds[rowind+1]-1)]
+        xcol = x[xinds[rowind]:(xinds[rowind+1]-1)]
         for colind in 1:rows[rowind]
             mapind +=1
             mul!(y[yinds[colind]:yinds[colind+1]-1], adjoint(maps[mapind]), xcol, 1, 1)
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -172,6 +172,7 @@ function isblocksquare(A::BlockMap)
     return all(r -> r == N, rows)
 end
 
+# the following rules are sufficient but not necessary
 function LinearAlgebra.issymmetric(A::BlockMap)
     isblocksquare(A) ? N = length(A.rows) : return false
     maps = A.maps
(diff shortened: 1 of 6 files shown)


dkarrasch:
I wasn't aware of `Dims`, that looks indeed as what we need. Thanks!

The `cumsum` thing is actually nothing but recursion, no iterator. I call it `sumshift` to indicate that in each recursion step, you shift the first element from the second tuple over to the first, while adding it the last entry of the first tuple. Since at compile time, you know the length of the tuples and their element types, the type of the result is perfectly type inferable, and the program will know what to expect, except for the precise values.

I guess (but as I said, I was surprised myself), that this allows for subsequent optimizations by the compiler, which reduces much of the overhead. Perhaps, in Matlab you wouldn't gain anything over plain cumsum, so that's why nobody would care to write a recursive function in that place.

dkarrasch:
Yes, `Dims` works, but I wonder if it creates a bit of confusion. It is meant to "represent the dimensions of an `AbstractArray`.", as the docs say. So, in a `Matrix`/`LinearMap` context, you would expect to see two-tuples, and not higher-order tuples, which represent the first index of blocks...

JeffFessler:
Oh right, I completely agree.

dkarrasch:
Hey, look at this!
```julia
using LinearMaps, LinearAlgebra, BenchmarkTools
A = LinearMap(rand(100, 100))
L = hvcat((3, 3), A, I, I, A, I, I)
x = rand(size(L, 2))
y = rand(size(L, 1))
@btime mul!($y, $L, $x)
M = Matrix(L)
@btime mul!($y, $M, $x)
```
gives
```julia
6.269 μs (32 allocations: 1.34 KiB)
7.166 μs (0 allocations: 0 bytes)
```
The larger the blocks, the more it pays off to have the optimized uniform scaling multiplication and to really use the block structure. In that sense, this the clearly the way to go, and not to put everything in one big matrix via `BlockArrays.jl`.

Generally, there is probably still much room for improvement. In the multiplication, there are the two calls to `firstrowindices` and to `firstcolindices`. Both are inferable, but allocate. I'm having fun trying to optimize this, but couldn't find a better solution yet. Attempts like
```julia
function firstrowindices(A::LinearMaps.BlockMap)
    as, rows = A.maps, A.rows

    index = 1
    cumsize = 1
    return tuple(1, ntuple(length(rows)-1) do i
        cumsize += size(as[index], 1)
        index += rows[i]
        return cumsize
    end...)
end
```
and
```julia
sumshift(a::Tuple{Vararg{Int}}, b::Int) = tuple(a..., last(a) + b)
colindices(As::Tuple{Vararg{LinearMap}}) = mapreduce(a -> size(a, 2), sumshift, As; init=(1,))
```
don't seem to improve anything. Anyway, this is in pretty good shape already, and I'm excited if we manage to further improve it.

dkarrasch:
Haha, I just found @Jutho's TupleTools.jl and the cumsum function there. Maybe we could use some of those functions (I like the `getindex` for instance), but my first attempts again don't improve anything. I guess I'll call it a day on a Sunday, but remain excited if @Jutho can squeeze out some more performance.

dkarrasch:
Maybe it's as easy as that: compute the block indices at construction and save them in a field of the `BlockMap` struct? There is no reason to repeat this computation at every multiplication, because this is tied to the BlockMap. This should take off lots of the pressure to optimize out these steps.

dkarrasch:
Alright, in my little 10x10 benchmark from the OP, I am now down on this:
```julia
422.231 ns (8 allocations: 480 bytes)
  97.380 ns (0 allocations: 0 bytes)
```
This is faster by almost factor 8 than current master. I figured that since now the range computations are one-time computations, there is little if any benefit from working with tuples and recursive code. Instead, I use nice concrete `Vector`s of either `Int`s or `UnitRange{Int}`. I have peppered the multiplication with `@inbounds` after checking that the total size of `x` and `y` match the corresponding dimensions of `A`. The internal subindexing is then "known" to be correct because this is how we computed the index ranges at construction of `A`.

I believe the code is also pretty clean by now. What is still missing is an attempt to automatically identify symmetry, and potentially some nice `show` method.

Jutho:
Let me know when this is ready for a review.

dkarrasch:
Yes, I'm almost done with the symmetry stuff, then I'll let go.

dkarrasch:
After tests pass, this is ready for review. To summarize:

The new philosophy is to compute all ranges at construction. This reduces the multiplication overhead as described above. I couldn't get anything better with tuples and recursion here. Early on, I moved the dimension checks into the promotion part, as @Jutho suggested in the first PR. Then, symmetry checks now work, at least for block symmetric/Hermitian operators. So this should probably get a comment like "sufficient condition" or so. Finally, I have added a 5-arg `mul!` for `UniformScalingMap`s, to avoid any allocation there.

JeffFessler:
I agree that precomputing the ranges at construction with simple code is the way to go.
I should have thought of suggesting this earlier because I do something analogous (but messier) in my matlab toolbox
`https://github.com/JeffFessler/mirt/blob/master/systems/%40fatrix2/private/fatrix2_vertcat.m`

src/blockmap.jl
```diff
@@ -57,9 +69,7 @@ function Base.hcat(As::Union{LinearMap,UniformScaling}...)
     end
     nrows == 0 && throw(ArgumentError("hcat of only UniformScaling-like objects cannot determine the linear map size"))
```
Jutho: I still feel uncomfortable that this changes Base behaviour, also, the error message is somewhat different. One way out would be to call this function `_hcat` and then define
```julia
for k = 1:8 # is 8 sufficient?
      Is = ntuple(n->:($(Symbol(:A,n))::UniformScaling), Val(k-1))
      L = :($(Symbol(:A,k))::LinearMap)
      args = ntuple(n->Symbol(:A,n), Val(k))

      @eval Base.hcat($(Is...), $L, As::Union{LinearMap,UniformScaling}...) = _hcat($(args...), As...)
      @eval Base.vcat($(Is...), $L, As::Union{LinearMap,UniformScaling}...) = _vcat($(args...), As...)
      @eval Base.hvcat(rows::Dims, $(Is...), $L, As::Union{LinearMap,UniformScaling}...) = _hvcat(rows, $(args...), As...)
end
```

src/blockmap.jl
```diff
@@ -141,51 +149,58 @@ function Base.hvcat(rows::NTuple{nr,Int}, As::Union{LinearMap,UniformScaling}...
...
-    (promote_to_lmaps_(n[k], A), promote_to_lmaps_(n[k+1], B), promote_to_lmaps(n, k+2, Cs...)...)
+promote_to_lmaps_(n::Int, dim, J::UniformScaling) = UniformScalingMap(J.λ, n)
+promote_to_lmaps_(n::Int, dim, A::LinearMap) = (check_dim(A, dim, n); A)
+promote_to_lmaps(n, k, dim) = ()
+promote_to_lmaps(n, k, dim, A) = (promote_to_lmaps_(n[k], dim, A),)
+promote_to_lmaps(n, k, dim, A, B) = (promote_to_lmaps_(n[k], dim, A), promote_to_lmaps_(n[k+1], dim, B))
+promote_to_lmaps(n, k, dim, A, B, C) =
+    (promote_to_lmaps_(n[k], dim, A), promote_to_lmaps_(n[k+1], dim, B), promote_to_lmaps_(n[k+2], dim, C))
+promote_to_lmaps(n, k, dim, A, B, Cs...) =
+    (promote_to_lmaps_(n[k], dim, A), promote_to_lmaps_(n[k+1], dim, B), promote_to_lmaps(n, k+2, dim, Cs...)...)
```
Jutho: This last method also captures the previous two, no? Adding `@inline` in front should make sure that there is no performance difference with or without having the former two.

src/blockmap.jl
```diff
@@ -141,51 +149,58 @@ function Base.hvcat(rows::NTuple{nr,Int}, As::Union{LinearMap,UniformScaling}...
         end
     end
 
-    return BlockMap{T}(promote_to_lmaps(n, 1, As...), rows)
+    return BlockMap{T}(promote_to_lmaps(n, 1, 1, As...), rows)
```
Jutho: is the `if needcols` necessary? Is it not anyway worthwhile to check consistent total column size.

Also, since Github doesn't allow me to comment on old lines which have not been edited in this PR:
```julia
nci = nc ÷ rows[i]
nci * rows[i] != nc && throw(...)
```
can probably be captured by `nci, r = divrem(nc, rows[i]); r != 0 && throw(...)`

Jutho:
Looks great; a few suggestions, but otherwise ready to merge I think?

dkarrasch:
Many thanks to all for reviewing and approving. @Jutho If I don't get to it by tomorrow, please feel free to take over and push it over the finish line. I'll be offline for more than 2 weeks then.

src/blockmap.jl
dkarrasch: I've copied this as is from Base.uniformscaling.jl. I think it might be necessary if you do have a row with only `UniformScaling`s. Honestly, I haven't even tried to parse the logic here, because I knew it was tested in Base, and does serve the exact same purpose as we want it here.

Regarding the old line, you should make a PR to Julia Base 😛. Anyway, I've included it in the PR.

src/blockmap.jl
dkarrasch: Yes, I commented out the previous two lines and `@inline`d the `A, B, Cs...` one, and it seems to work fine with one, two, three and more linear maps. So I removed the previous two lines.

src/blockmap.jl
```diff
@@ -203,65 +218,65 @@ LinearAlgebra.adjoint(A::BlockMap)  = AdjointMap(A)
...
-        yrow = @views y[yinds[rowind]:(yinds[rowind+1]-1)]
+    @views @inbounds for rowind in 1:length(rows)
+        yrow = y[yinds[rowind]]
         mapind += 1
-        A_mul_B!(yrow, maps[mapind], x[xinds[1]:xinds[2]-1])
+        A_mul_B!(yrow, maps[mapind], x[xinds[mapind]])
         for colind in 2:rows[rowind]
             mapind +=1
-            mul!(yrow, maps[mapind], x[xinds[colind]:xinds[colind+1]-1], 1, 1)
+            mul!(yrow, maps[mapind], x[xinds[mapind]], 1, 1)
```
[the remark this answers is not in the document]
dkarrasch: Yeah, it works with `true`s, so I put those.

src/blockmap.jl
dkarrasch: The change in the error message was a left-over from earlier attempts (and `vcat` had the `hcat` message). As a quick fix, I plugged in the same error message as in Base. As an intermediate step, I'll make a commit with your other suggestions and this quick fix. If I find the time, I'll try to use your above code, but if anything fails, I'll leave it at that for the moment.

src/blockmap.jl
dkarrasch: Oh, and I guess that change will actually help us test the method with no linear map argument. 🎉

src/blockmap.jl
dkarrasch: Judging from coverage test results, we do seem to have one test where the `needcols` branch is run. I strongly suspect it's this one:
```julia
@test [I; LA LA LA] isa LinearMaps.BlockMap{elty}
```
in line 541.

Jutho: I guess I misexpressed myself. The block inside the `if needcols` is definitely necessary. I was just wondering if we needed the `if` itself. This block can also be ran, and then it just checks whether total column size is consistent. If there is a row of `I`s, it will also compute their size.

dkarrasch: Cool. That turns out to be important. I had a test, which was supposed to be throwing, and it did throw, but not because of our code, but because of `LinearAlgebra`. Removing the `if`-branch, now, as you say, we check for consistent total column size, and the code throws as expected.

src/blockmap.jl
dkarrasch: Copy, paste, and it works... not my code, obviously. 🤣

dkarrasch:
Alright, I think this PR now has everything: updated README, updated version, tests on all Julia versions that are maintained, and all issues/comments resolved. I'll wait a few minutes for all tests to pass (they do pass locally), then merge, then open a registrator issue.

Jutho:
One question / idea: With the new way of dealing with `hcat` and friends, we could actually also support `AbstractMatrix` types which are not wrapped as `LinearMap`. As soon as there is one `LinearMap` in the list, our methods would take over. But probably this needs a small adaptation to then wrap these arguments as LinearMap, I think only in the `promote_to_lmaps` step.

Jutho:
But maybe that's an idea for a next PR. Anyway, this is definitely good to merge in my opinion.

dkarrasch:
Yes, that was my very early dream, about which I have forgotten with all the struggle in between. Wouldn't it simply require
```julia
for k in 1:8 # is 8 sufficient?
      Is = ntuple(n->:($(Symbol(:A,n))::Union{UniformScaling,AbstractMatrix}), Val(k-1))
      L = :($(Symbol(:A,k))::LinearMap)
      args = ntuple(n->Symbol(:A,n), Val(k))

      @eval Base.hcat($(Is...), $L, As::Union{LinearMap,UniformScaling,AbstractMatrix}...) = _hcat($(args...), As...)
      @eval Base.vcat($(Is...), $L, As::Union{LinearMap,UniformScaling,AbstractMatrix}...) = _vcat($(args...), As...)
      @eval Base.hvcat(rows::Tuple{Vararg{Int}}, $(Is...), $L, As::Union{LinearMap,UniformScaling,AbstractMatrix}...) = _hvcat(rows, $(args...), As...)
end
```
? Maybe we should increase the number 8 then, to allow for more than 8 matrices/scalings first...

PR #59, opened 2019-08-24
Fix index in BlockMap adjoint/transpose multiply

### LINKED ISSUE #58: hvcat indexing issue with misaligned blocks
While doing more thorough testing of `LinearMapsAA` I uncovered this issue with the new blockmap features (that are otherwise working great):

```
A = LinearMap(ones(3,2))
B = [I A; A I]
Matrix(B) # works fine
Matrix(B') # fails with error DimensionMismatch("mul!")
```

Now, one could argue that the way `B` is constructed in this MWE, with blocks that do not quite line up, is not the way we write block matrices in math, and I agree.  But that type of `hvcat` works fine with ordinary arrays, because the two block rows have the same number of columns.  I'm not sure it is worth trying to support multiplication with such constructs.  Perhaps `LinearMaps.hvcat` could just throw an error saying that such constructions are unsupported, e.g., by checking that all blocks in a given column have the same number of columns.

JeffFessler:
Fixes #58.

runtests needs additional tests such as:

```
A = LinearMap(ones(3,2))
B = [I A; A I]
@test Matrix(B') == Matrix(B)'
@test Matrix(transpose(B)) == Matrix(B)'
@test Matrix(adjoint(B)) == Matrix(B)'
```

### DIFF
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -244,19 +244,19 @@ function At_mul_B!(y::AbstractVector, A::BlockMap, x::AbstractVector)
     @boundscheck (n == length(y) && m == length(x)) || throw(DimensionMismatch("At_mul_B!"))
     maps, rows, xinds, yinds = A.maps, A.rows, A.rowranges, A.colranges
     mapind = 0
-    # first block row (rowind = 1), fill all of y
+    # first block row (rowind = 1) of A, meaning first block column of A', fill all of y
     @views @inbounds begin
         xcol = x[xinds[1]]
         for colind in 1:rows[1]
             mapind +=1
-            A_mul_B!(y[yinds[colind]], transpose(maps[mapind]), xcol)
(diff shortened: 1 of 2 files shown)


JeffFessler:
For future reference: `hvcat` works row-wise and there is no corresponding `vhcat` but the transpose of a `hvcat` is akin to what a hypothetical `vhcat` would make.
All of the existing nice indexing for row and column pieces works fine for the transpose.

JeffFessler:
The failing tests seem to be 1.3 related.  I tested only with 1.1 locally before submitting the PR.

dkarrasch:
Awesome, nice catch! Could you please add the proposed test at the end of the `hvcat` test section? The test failure is indeed related to the update by the 5-arg `mul!`. We can merge this with the test, and afterwards take care of adding version restrictions.

JeffFessler:
Done.  I think...

PR #61, opened 2019-08-25
add Kronecker products for LinearMaps

dkarrasch:
I realized that part of @JeffFessler's motivation for having `LinearMaps <: AbstractArray` is to use them in combination with [Kronecker.jl](https://github.com/MichielStock/Kronecker.jl). I thought I'll lower the pressure and provide a Kronecker product for `LinearMap`s without the need for `getindex`. 😛 

The implementation uses the well-known vec trick, and since the Kronecker product is associative, one only needs to take care of the product of two linear maps. I couldn't find any trick for "higher-order" Kronecker products. Note that this is all lazy, and all linear map types are, as usual, inferred. This works out of the box with wrapped maps of dense/sparse/structured matrices, or function maps, etc.


dkarrasch:
> One could easily add a method for the product of two Kronecker maps,
> returning yet another Kron. map.

I saw that in Wikipedia, but I thought that works only when the matrix products exists, but they don't have to...?

dkarrasch:
Hooray, I managed to add promotion of matrices (as long as there is one `LinearMap` among the first 8 arguments of `kron`. That took me a great while, to figure out the metaprogramming issues. I'm not sure I understood what turned out to be correct, but anyway.

dkarrasch:
In terms of numerical effort, I assume it would be beneficial to use that mixed-property? Could be that this would be a second "good" use case for the union splitting mechanism.

The promotion thing is really "simple". All it does is to take the matrix arguments and wraps them in a `LinearMap`. Finally, it simply calls `kron` again on the same list (up to to lmap-promotion). Ultimately, how is `kron(A, B, C)` handled`? Well, by Julia `Base`. It iteratively calls `kron(kron(A, B), C)`, but this time, everytime `kron` is called, it is called with only two arguments, each a `LinearMap`, so the method from two lines above gets called. The vec trick applies consecutively, first to `(A x B) x C` (whatever `A x B` is at that point), so it's `transpose(C)*X*(A x B)`, and when you multiply `A x B` with some `x`, you again apply the vec trick.

I'll add some comments to the code.

dkarrasch:
I just did a quick check, and in fact it's beneficial to use the mixed-trick. I'll put it in the PR, then we discuss the stability issue.

dkarrasch:
I thought I'll post a little comparison here for general entertainment. I'm sure I have used `LazyArrays.jl` incorrectly and I'm happy to fix the code below, but here is what I tested:
```julia
using BenchmarkTools, Test
using LinearAlgebra, SparseArrays
using LinearMaps, Kronecker, LazyArrays
n = 100
x = rand(n*n)
y = zeros(n*n)

@info "LinearMaps.jl"
J = LinearMaps.UniformScalingMap(true, n)
E = spdiagm(-1 => trues(n-1))
D = E + E' - 2I
A1 = kron(D, J) + kron(J, D)
@btime $A1 * $x

@info "Kronecker.jl"
A2 = kronecker(D, J)
A2´ = kronecker(J, D)
@btime $A2 * $x + $A2´ * $x

@info "LazyArrays.jl"
A3 = Kron(D, J)
A3´ = Kron(J, D)
@btime $A3 * $x + $A3´ * $x

@info "Base"
A4 = kron(D, J) + kron(J, D)
@btime $A4 * $x

@test sparse(A1) == A4
```
and here is what I got:
```julia
[ Info: LinearMaps.jl
  956.586 μs (726 allocations: 521.30 KiB)
[ Info: Kronecker.jl
  5.115 ms (64 allocations: 1.35 MiB)
[ Info: LazyArrays.jl
  5.771 s (6 allocations: 234.61 KiB)
[ Info: Base
  55.835 μs (2 allocations: 78.20 KiB)
```
This is the discrete Laplace operator. Note that with `Kronecker.jl` and `LazyArrays.jl` the multiplication needs to be done in two steps: if you add the two Kronecker products, you'll get a dense matrix, and, of course, performance is horrible. `Base` is so fast because the Kronecker product of sparse matrices is sparse again, but `Kronecker.jl` has shown that for dense matrices performance is much worse compared to `kronecker`.

One thing I noted is that there doesn't seem to be a "lazy" version of a `UniformScaling` with known size. It seems like you need to construct a sparse or diagonal matrix, with obvious redundant memory consumption. @Jutho So we may wish to consider exporting our `UniformScalingMap`, for instance, for constructions like the above.

JeffFessler:
Nice timing results!  I didn't know about `LazyArrays.jl` before.
On the plus side, `LazyArrays` has `getindex` and `setindex` and subtypes of `AbstractArray` :-)
The utf8 in `A2´` looks so much like `A2'` that it confused me at first btw!

dkarrasch:
Yeah, I didn't know which other symbol to use so I typed some random symbol. 😛 Well, and yes, the array type behavior is what distinguishes those packages from ours here.

Some day you'll have to explain me why you can't live without `getindex` and the `AbstractArray` subtype. Until that day, I'll try to provide every other feature that you need here. I'm thinking about adding a `BlockDiagonalMap` type, the corresponding thing to what I saw in `MIRT.jl`.

dkarrasch:
Regarding the minimal storage but known-size uniform scaling object, I just found this discussion here: https://github.com/JuliaLang/julia/pull/30298

The use of `I(n)` in Kronecker products seems to be indeed a very classic case. But I just realized that since `UniformScaling`s are not `AbstractMatrix`es, we could safely define
```julia
LinearMap(J::UniformScaling, n::Int) = UniformScalingMap(J.λ, n)
```
and then we don't need to export `UniformScalingMap`s.

dkarrasch:
Thanks, and mat-vec-multiplication with Kronecker products is also faster now. I reuse intermediate vectors (called `temp1` and `temp2`) when applying the vec trick. The Kronecker sum requires a bit of work on conversion of operators to matrices. We should add a couple of specialized conversion rules, especially when operators are made up of matrices. But with that, I'm going to wait until #66 and #68 are finished and merged.

dkarrasch:
This is still WIP, basically it needs some more tests, but I wanted to quickly share some impressive results, obtained on Julia v1.3 with the 5-arg `mul!`.

```julia
using BenchmarkTools, Test
using LinearAlgebra, SparseArrays
using LinearMaps, Kronecker#, LazyArrays
n = 100
x = rand(n*n)
y = zeros(n*n)

@info "LinearMaps.jl"
J = LinearMap(I, n)
E = spdiagm(-1 => trues(n-1))
D = E + E' - 2I
L = kron(D, J) + kron(J, D)
@btime $L * $x # 89.378 μs (27 allocations: 313.48 KiB)

@info "Kronecker.jl"
J = Kronecker.Eye(n)
A2 = kronecker(D, J)
A2´ = kronecker(J, D)
@btime ($y = $A2 * $x; $y.+= $A2´ * $x) # 89.393 μs (17 allocations: 313.20 KiB)

@info "Base"
J = I(n)
A4 = kron(D, J) + kron(J, D)
@btime $A4 * $x # 57.114 μs (2 allocations: 78.20 KiB)
```
We are getting really close to the super-sparse matrix multiplication. Interestingly, the operator is nothing but a Kronecker sum.
```julia
J = LinearMap(I, n)
E = spdiagm(-1 => trues(n-1)); D = E + E' - 2I; DL = LinearMap(D)
Δ₂ = LinearMaps.kronsum(DL, DL)
Δ₃ = kroneckersum(D, D)
@btime $Δ₂ * $x # 67.537 μs (7 allocations: 78.41 KiB)
@btime $Δ₃ * $x # 67.327 μs (7 allocations: 78.41 KiB)
```
So, we are able to fully leverage all `LinearAlgebra` functionality in the case of matrix-based linear maps!

dkarrasch:
This has grown big, as usual, but I think it's ready for review. For Kronecker products of matrix-based linear maps to work fast, there were quite some adjustments needed for `MatrixMap`s and `UniformScaling`s. One interesting aspect is that one sometimes cannot avoid intermediate matrix results. Obviously, when these are computed from matrices, you don't want to generate these by applying them to the canonical basis vectors. As demonstrated above, there is also quite some benefit from the 5-arg `mul!` implementation. I'd suggest to finalize and merge #66 before this one, and afterwards make a minor release.

src/uniformscalingmap.jl
```diff
@@ -14,10 +14,17 @@ LinearAlgebra.issymmetric(::UniformScalingMap) = true
 LinearAlgebra.ishermitian(A::UniformScalingMap) = isreal(A)
 LinearAlgebra.isposdef(A::UniformScalingMap) = isposdef(A.λ)
 
+# comparison of UniformScalingMap objects, sufficient but not necessary
+Base.:(==)(A::UniformScalingMap, B::UniformScalingMap) = (eltype(A) == eltype(B) && A.λ == B.λ && A.M == B.M)
```
Jutho: I am starting to doubt whether we should compare `eltype` in `==`. Surely `Number` and `AbstractArray` instances can be equal without having the same `eltype`.

src/kronecker.jl
```diff
@@ -0,0 +1,259 @@
...
+Construct a `KroneckerMap <: LinearMap` object, a (lazy) representation of the
+Kronecker product of two `LinearMap`s. One of the two factors can be an `AbstractMatrix`
+object, which then is promoted to `LinearMap` automatically. To avoid fallback to
+the generic [`Base.kron`](@ref), there must be a `LinearMap` object among the
+first 8 arguments in usage like `kron(A, B, Cs...)`.
+
+Note: If `A`, `B`, `C` and `D` are linear maps of such size that one can form the
+matrix products `A*C` and `B*D`, then the mixed-product property `(A⊗B)*(C⊗D)`
+holds. Upon multiplication of Kronecker products, this rule is checked for
+applicability, which leads to type-instability with a union of two types.
```
Jutho: I should probably read the discussion in this thread, but could you give a quick summary as to why this advantageous over just using `CompositeMap` for the product?

dkarrasch: Ah, that is a left-over from earlier approaches. Applying the mixed-product rule is advantageous because every Kronecker product allocates a `zero` vector on multiplication, so when you can replace a product of two (or more) Kronecker products by one Kronecker product of products, this saves allocation effort, and maybe further reduces the overall complexity. But the product rule is now implemented in `mul!`, and therefore type stable! I'll remove that sentence from the docstring.

src/kronecker.jl
```diff
@@ -0,0 +1,258 @@
...
+    X = LinearMap(reshape(x, (size(B, 2), size(A, 2))); issymmetric=false, ishermitian=false, isposdef=false)
+    _kronmul!(y, B, X, transpose(A), T)
+    return y
+end
+# mixed-product rule, prefer the right if possible
+# (A₁ ⊗ A₂ ⊗ ... ⊗ Aᵣ) * (B₁ ⊗ B₂ ⊗ ... ⊗ Bᵣ) = (A₁B₁) ⊗ (A₂B₂) ⊗ ... ⊗ (AᵣBᵣ)
+function A_mul_B!(y::AbstractVector, L::CompositeMap{<:Any,<:Tuple{KroneckerMap,KroneckerMap}}, x::AbstractVector)
+    B, A = L.maps
+    if length(A.maps) == length(B.maps) && all(M -> check_dim_mul(M[1], M[2]), zip(A.maps, B.maps))
+        A_mul_B!(y, kron(map(prod, zip(A.maps, B.maps))...), x)
```
Jutho: You could probably also use `map(*, A.maps, B.maps)`, which is maybe somewhat more intuitive?

src/kronecker.jl
```diff
@@ -0,0 +1,258 @@
...
+
+julia> Δ₂ = LinearMaps.kronsum(D, D); # same operator as Kronecker sum
+
+julia> Matrix(Δ₁) == Matrix(Δ₂)
+true
+```
+"""
+kronsum(A::LinearMap{TA}, B::LinearMap{TB}) where {TA,TB} = KroneckerSumMap{promote_type(TA,TB)}((A, B))
+kronsum(A::LinearMap, B::LinearMap, C::LinearMap, Ds::LinearMap...) = kronsum(A, kronsum(B, C, Ds...))
+# promote AbstractMatrix arguments to LinearMaps, then take LinearMap-Kronecker sum
```
Jutho: As there is no `kronsum` in `Base`, this is probably not necessary. Though there is no problem in keeping it for consistency, and for the days when this function exists in Base :-)

src/kronecker.jl
```diff
@@ -0,0 +1,258 @@
...
+Base.:(==)(A::KroneckerSumMap, B::KroneckerSumMap) = (eltype(A) == eltype(B) && A.maps == B.maps)
+
+function LinearMaps.A_mul_B!(y::AbstractVector, L::KroneckerSumMap, x::AbstractVector)
+    A, B = L.maps
+    ma, na = size(A)
+    mb, nb = size(B)
+    (length(y) == size(L, 1) && length(x) == size(L, 2)) || throw(DimensionMismatch("kronecker product"))
+    X = reshape(x, (nb, na))
+    Y = reshape(y, (nb, na))
+    mul!(Y, X, convert(AbstractMatrix, transpose(A)))
```
Jutho: How about `transpose(convert(AbstractMatrix, A))`, to make use of the methods to multiply with transpose/adjoint matrices in Base. I was thinking about the case where `A` is a `MatrixMap`.

However, I just checked. There is no `convert(AbstractMatrix, ::LinearMap)`, only `convert(Matrix, ::LinearMap)`. So I assume this line of code is not  Furthermore, `convert(Matrix, ::MatrixMap)` does actually call the generic method instead of just returning the matrix it wraps. 

Never mind, I just noticed that you also fixed all of this in this PR :-).

src/kronecker.jl
```diff
@@ -0,0 +1,258 @@
...
+"""
+    kron(A::LinearMap, k::Int)
+
+Construct a lazy representation of the `k`-th Kronecker power `A ⊗ A ⊗ ... ⊗ A`
+for `k≥2`. This function is currently not type-stable!
+"""
+function Base.kron(A::LinearMap, k::Int)
+    k > 1 || throw(ArgumentError("the Kronecker power is only defined for exponents larger than 1, got $k"))
+    return kron(Base.fill_to_length((), A, Val(k))...)
+end
```
Jutho: I would assume that type stability is ok when `k` is a constant in the source code, i.e. a literal integer.

However, where is this function/definition coming from. Also its position in the source code is very strange, since this is about Kronecker product, not kronecker sum, but it also seems like something that should not be a method of `kron`. This seems analoguous to defining `*(a::Number, k::Int) = a^k`?

src/kronecker.jl
dkarrasch: The conversion rules are most likely not complete. They cover what we used to have and matrix conversion that we need anyway in the Kronecker multiplication. I just thought that with block maps and 5-arg mul!, there is additional use of our package with only matrices, for which conversion doesn't necessarily require multiplying with tons of canonical vectors. So I collected whatever came to my mind would be cheaply obtained directly from matrices. This could be extended as required.

src/kronecker.jl
dkarrasch: This is a convenience functionality that I "copied" from Kronecker.jl. There, however, additional types (for the Kronecker power and the Kronecker sum power) are introduced, for which then special mixed-product/multiplication rules are introduced etc. Since everything is completely lazy here (and there, actually), I thought there is no reason to introduce new types, but to "blow up" `A` to `(A, A, ..., A)` and to pass it to `kron`. We could change the function name and make it, say, `kronpow` or so, but this adds to the number of exported functions, which I wanted to keep small. But I really don't have strongly feelings either way.

In a sense, it is analogous to what you're saying, except that we don't have an alternative notation for the right hand side in the Kronecker context.

src/kronecker.jl
dkarrasch: What do you mean? `KroneckerSum` currently expects `LinearMap` objects, or do you think we could relax this? The question is whether `mul!` continues to work for "mixed" `maps` tuples, or at least requires only mild modifications.

src/kronecker.jl
Jutho: I still think that this is in violation of what functions and different methods thereof are supposed to represent, according to the Julia guidelines/common practice. A function should have one specific functionality, albeit this being implemented in different ways for different types. These are two distinct types of functionality, kronecker product and kronecker power. In particular, because there is a perfectly valid interpretation to the kronecker product of a matrix and a number. In fact, in Base (actually, standard library LinearAlgebra)
`kron( ::Matrix, ::Number)` is implemented, and does what it means mathematically, i.e. `M ⊗ s = M * s`, with `M` matrix and `s` a scalar.  

So I would indeed suggest a new method. Another possibility is some unicode hackery and exporting `⊗`, though technically that is also a new function.

```julia
julia> struct KronWrap{T<:Integer}
               p::T
       end

julia> ⊗(k::Integer) = KronWrap(k)
⊗ (generic function with 1 method)

julia> ⊗(a, b, c...) = kron(a, b, c...)
⊗ (generic function with 2 methods)

julia> Base.:^(a::Matrix, p::KronWrap) = kron(ntuple(n->a, p.p)...)

julia> a = randn(2,2)
2×2 Array{Float64,2}:
 -1.16944    0.141176
  0.590445  -1.12108 

julia> a^⊗(3)
8×8 Array{Float64,2}:
 -1.59933    0.193071    0.193071   -0.0233076   0.193071   -0.0233076  -0.0233076   0.00281371
  0.807491  -1.53319    -0.0974806   0.185087   -0.0974806   0.185087    0.0117679  -0.0223438 
  0.807491  -0.0974806  -1.53319     0.185087   -0.0974806   0.0117679   0.185087   -0.0223438 
 -0.407697   0.774098    0.774098   -1.46979     0.0492173  -0.0934494  -0.0934494   0.177433  
  0.807491  -0.0974806  -0.0974806   0.0117679  -1.53319     0.185087    0.185087   -0.0223438 
 -0.407697   0.774098    0.0492173  -0.0934494   0.774098   -1.46979    -0.0934494   0.177433  
 -0.407697   0.0492173   0.774098   -0.0934494   0.774098   -0.0934494  -1.46979     0.177433  
  0.205844  -0.390837   -0.390837    0.742086   -0.390837    0.742086    0.742086   -1.40901   

julia> a ⊗ a ⊗ a
8×8 Array{Float64,2}:
 -1.59933    0.193071    0.193071   -0.0233076   0.193071   -0.0233076  -0.0233076   0.00281371
  0.807491  -1.53319    -0.0974806   0.185087   -0.0974806   0.185087    0.0117679  -0.0223438 
  0.807491  -0.0974806  -1.53319     0.185087   -0.0974806   0.0117679   0.185087   -0.0223438 
 -0.407697   0.774098    0.774098   -1.46979     0.0492173  -0.0934494  -0.0934494   0.177433  
  0.807491  -0.0974806  -0.0974806   0.0117679  -1.53319     0.185087    0.185087   -0.0223438 
 -0.407697   0.774098    0.0492173  -0.0934494   0.774098   -1.46979    -0.0934494   0.177433  
 -0.407697   0.0492173   0.774098   -0.0934494   0.774098   -0.0934494  -1.46979     0.177433  
  0.205844  -0.390837   -0.390837    0.742086   -0.390837    0.742086    0.742086   -1.40901   
```

dkarrasch: Yeah, that's cool. I would call that struct `KronPower`, to indicate that `⊗(3)` is to be understood as a Kronecker power, in contrast to a regular multiplicative power. Moreover, we can use the same trick with the Kronecker sum power, via exporting `⊕` and introducing `KronSumPower`.

Jutho: Indeed. I have to admit that I had never heard of a kronecker sum before (even though I use/encounter this structure often), but find it very confusing that it is the same symbol as a direct sum, which is clearly very different.

dkarrasch: Hm, I just did the change. Do you think there will be any conflict with the direct sum if we start exporting `\oplus`? It seems pretty common in the "Kronecker community" to use it for the Kronecker sum.

Jutho: I did not even know there was a "Kronecker community" :-). Good documentation is key, I guess. There is no dedicated direct sum in Base either, and it just seems to fit in the wider context of block matrices, so I think it's fine to have `\oplus` be Kronecker sum.

dkarrasch: > I did not even know there was a "Kronecker community" :-).

Maybe I should've used "matrix product community". 😉 There are many other similar matrix products people seem to use in, for instance, signal processing and other fields, see https://en.wikipedia.org/wiki/Kronecker_product#Related_matrix_operations

One thing to think about is how we want to (at some point) construct block diagonal maps. For that purpose, \oplus might be handy as well.

What I like a lot about `LinearMaps.jl` pre-Kronecker product is that we only export `LinearMap`, and everything else is overloaded from Base/LinearAlgebra, in contrast to what is done in other similar packages. That allows for nice generic programming, without the need to use dedicated functions which work for LinearMaps exclusively. But I guess we may want to overload `cat(As...; dims=(1,2))` for block diagonal concatenation, so using oplus for Kronecker sum (with clear documentation) should be indeed fine.

JeffFessler: `SparseArrays.jl` uses `blockdiag` which seems OK, though `Diagonal(As...)` would be a logical extension of `Diagonal(v)` and seems more natural/readable than `cat(...)`.
Right now my PR #65 uses `block_diag` because I was getting error messages about conflicts with `SparseArrays.blockdiag` that I could not resolve, but that's off topic.
I've never seen `\oplus` used for block diagonals.
I have no concerns about a new export because one could do `using LinearMaps: LinearMap` to get just the desired functions.

dkarrasch:
Brief update: I made the Kronecker (sum) power a type parameter, so this way everything is inferable. This puts a little overhead onto the compiler, but I guess this is not going to be used with many different powers.

Jutho:
Hi @dkarrasch , could you remind me again which open PRs are finalised and ready for review; I keep forgetting and was very busy the last weeks.

dkarrasch:
It would be good to get this one done here.

dkarrasch:
Sorry for messing around. When I'm bored, I'm starting to do micro-optimization. 😂  I've now optimized dimension checks, and included the option to skip them (as usual) via `@inbounds`. These checks always refer to whether the lengths of `y` and `x` are consistent with that of `A` (but not to one-based indexing). In many cases, this is guaranteed upfront, and there is no need to re-check this everytime. Accordingly, we may then wish to propagate that knowledge. I hope I have carefully distinguished when we need `Base.@propagate_inbounds` and when `@inline`.

Okay, I promise I'll leave this PR in peace for review. 🤞 But actually, the recent changes are just standard refactors of the previous code, no design changes or anything like that.

dkarrasch:
gentle bump

src/kronecker.jl
Jutho: My apologies for the late response. I meant that, since `kronsum` is not in base and does "belongs" to LinearMaps.jl, there is no type piracy in just defining
`kronsum(As::Vararg{Union{LinearMap,AbstractMatrix}})`

Jutho:
I tried to complete the review now but I am too tired; this time I will try not to forget.

src/kronecker.jl
dkarrasch: I see, and I agree. I simplified the construction accordingly. I doubt that `kronsum` is gonna become a `Base` function. Without this structural view, where you keep the factors, it doesn't make much sense to add it.

Jutho:
Generally looks good; I think this can be merged, maybe some small comments for consideration.

src/kronecker.jl
```diff
@@ -0,0 +1,285 @@
...
+
+"""
+    ⊗(k::Integer)
+
+Construct a lazy representation of the `k`-th Kronecker power `A^⊗(k) = A ⊗ A ⊗ ... ⊗ A`,
+where `A` can be an `AbstractMatrix` or a `LinearMap`.
+"""
+⊗(k::Integer) = KronPower(k)
+
+⊗(a, b, c...) = kron(a, b, c...)
```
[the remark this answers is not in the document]
dkarrasch: Cool, that's because we "own" `⊗`, right? I guess we can do the same with
```julia
Base.:(^)(A::Union{LinearMap,AbstractMatrix}, ::KronPower{p})
```

src/kronecker.jl
```diff
@@ -0,0 +1,285 @@
...
+
+Base.:(^)(A::Union{LinearMap,AbstractMatrix}, ::KronPower{p}) where {p} = kron(Base.fill_to_length((), convert_to_lmaps_(A), Val(p))...)
```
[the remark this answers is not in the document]
dkarrasch: I think I tried that, but it wasn't inferrable? I'll try again and see...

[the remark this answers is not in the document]
dkarrasch: I must have messed up something earlier. I was struggling with this quite a bit, but it's perfectly inferable. Not sure what my problem was before.

src/kronecker.jl
```diff
@@ -0,0 +1,277 @@
...
+        A_mul_B!(temp2, X, temp1)
+        A_mul_B!(y[((i-1)*mb+1):i*mb], B, temp2)
+        v[i] = zero(T)
+    end
+    return y
+end
+@inline function _kronmul!(y, B::Union{MatrixMap,UniformScalingMap}, X, At::Union{MatrixMap,UniformScalingMap}, T)
+    na, ma = size(At)
+    mb, nb = size(B)
+    if (nb + ma) * na < (ma + mb) * nb
```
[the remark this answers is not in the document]
dkarrasch: I took them from `Kronecker.jl`. I think this is to pick the order of (dense) matrix multiplications that requires less effort, but I admit that I never checked myself if these are the right conditions.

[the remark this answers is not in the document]
Jutho: The number of multiplications in `B*(X*At)` is `nb*na*ma + mb*nb*ma`, those of `(B*X)*At` is `mb*nb*na + mb*na*ma`, so if this is the determining factor the inequality should be
`nb*na*ma + mb*nb*ma < mb*nb*na + mb*na*ma`
or thus
`nb*ma*(na + mb) < mb*na*(nb+ma)`
Intermediate memory requirement, on the other hand, is the size for `X*At` is `nb*ma` or the size for `B*X` which is `mb*na`.
Not sure what the above inequality is.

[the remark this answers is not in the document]
dkarrasch: I just had the exact same considerations, and am running tests at this moment. 😄 

First of all, the inequality/branches is/are wrong in `Kronecker.jl`. I guess I'll simply go with the memory requirement.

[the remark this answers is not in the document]
dkarrasch: I'll go with the simple memory requirement condition. Counting flops can be misleading:
```julia
using BenchmarkTools, Test
using LinearAlgebra, SparseArrays
using LinearMaps, Kronecker

for (ma, na) in ((10, 501), (501, 10)), (mb, nb) in ((11, 499), (499, 11))
	A = rand(ma, na)
	B = rand(mb, nb)
	L = kron(LinearMap(A), B)
	K = kronecker(A, B)
	x = rand(size(L, 2))
	@show ma, na, mb, nb
	@info "take X*At path"
	@show nb*ma < mb*na
	@show (nb + ma) * na < (na + mb) * nb
	@show (nb+ma)*mb*na < (mb+na)*nb*ma
	@info "LinearMaps.jl"
	@btime $L * $x
	@info "Kronecker.jl"
	@btime $K * $x
end
```
For the combination `(ma, na, mb, nb) = (501, 10, 11, 499)` I get
```julia
nb * ma < mb * na = false
(nb + ma) * na < (na + mb) * nb = true
(nb + ma) * mb * na < (mb + na) * nb * ma = true
[ Info: LinearMaps.jl
  17.733 μs (12 allocations: 44.44 KiB)
[ Info: Kronecker.jl
  398.609 μs (8 allocations: 1.95 MiB)
```
clearly an unfortunate choice.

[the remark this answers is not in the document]
dkarrasch: To help the eye: the first condition is the intermediate memory allocation, the second is the current condition from `Kronecker.jl`, in whose adaptation I had a typo before, and the third is the total number of flops.

PR #63, opened 2019-08-28
make A*I return A

### LINKED ISSUE #57: Make LinearMap * I returns same LinearMap
Currently,

`*(A::LinearMap, I::UniformScaling{Bool})::CompositeMap`

That is, a new map is constructed. Could we simply return `A` in this case (and related cases involving multiplication/division by the identity?)

dkarrasch:
If we decide to do it, this closes #57.


dkarrasch:
> the typical use case is something like
> `B = A * (do_precon ? P : I)`
> in the preamble of some code, then after that there would be repeated use of `B*x` in some iterative code.

This is exactly the issue that I worry about: at compilation time, the code will not know of which type `B` is. As I said in the other PR, maybe this is exactly what union splitting is good for, so that "two versions" corresponding to the two possible outcome types are compiled, and eventually everything looks like it was type-stable. The other option is to have a so-called function barrier, so between the generation of `B` and the `B*x` there is a function call, at which point the type of `B` will be known, and a specialized method gets compiled. But maybe the call to `mul!` or `*` actually already does serve as a function barrier?

src/linearcombination.jl
```diff
@@ -37,6 +37,18 @@ function Base.:(+)(A₁::LinearMap, A₂::LinearMap)
     return LinearCombination{T}(tuple(A₁, A₂))
 end
 Base.:(-)(A₁::LinearMap, A₂::LinearMap) = +(A₁, -A₂)
+function Base.:(+)(O::ZeroMap, A::LinearMap)
+    size(O) == size(A) || throw(DimensionMismatch("+"))
+    T = promote_type(eltype(O), eltype(A))
```
Jutho: what is the point of computing `T` if then anyway `A` is returned?

src/linearcombination.jl
```diff
@@ -37,6 +37,18 @@ function Base.:(+)(A₁::LinearMap, A₂::LinearMap)
...
+    return A
+end
+Base.:(+)(A::LinearMap, O::ZeroMap) = +(O, A)
+function Base.:(+)(O₁::ZeroMap, O₂::ZeroMap)
+    size(O₁) == size(O₂) || throw(DimensionMismatch("+"))
+    T = promote_type(eltype(O₁), eltype(O₂))
```
Jutho: Same here

Jutho:
I am not too fond of the type instability. Another way to resolve this, is to special case the multiplication rule of `CompositeMap` when one of the maps is a `UniformScalingMap`. This is likely to be hard though.

Yet another solution is to have a `ScaledMap` which just encodes a linear map and a scalar factor. `ZeroMap` would be a special case of that. This functionality could also be added to `WrappedMap`, though that is maybe throwing together too much.

src/linearcombination.jl
dkarrasch: Good point. This situation here is supposed to occur when the `ZeroMap` was created out of a `CompositeMap` that had a `ZeroMap` factor. So that map does have its own eltype, which probably should not get lost here. So I guess what should be returned is actually `LinearMap{T}(A)`. At least that's what I meant to do originally, but then forgot about it in the next line.

PR #66, opened 2019-09-02
miscellaneous updates

dkarrasch:
* ~~add size(A, i)
  I realized we more often use `size(A, i)` instead of `size(A)`, and the interface requests to have `size(A)`, and `size(A, i)` computes by default `size(A)[i]`. As for the API, I think this is fine, but I felt we should have specialized methods for individual sizes whenever possible to save unnecessary computations. In all cases that effort is trivial compared to others, but still in some cases there is a little work to be done.~~
* ~~make FunctionMap adjoint/transpose-invariant
  We were handling adjoints/transposes of `FunctionMaps` a bit loose. We never checked at construction whether the adjoint is defined, and we never checked whether the adjoint is also in-place. ~~Since all information is available, I thought we may as well construct a new `FunctionMap`, and at that point guess again whether `fc` is also in-place or not. Also, this saves us one hierarchical level.~~ EDIT: That doesn't work well. Instead, I added some test. Strictly speaking, we still can define a `FunctionMap` where `f` is in-place, and `fc` isn't. This is still not caught a priori, but will yield an error at multiplication.~~
* require one-based indexing in multiplications that index into the vectors
  I think we were a bit too loose at that point, compared to our signature. Most linalg operations in Base are restricted to such vectors anyway.
* make `alpha` and `beta` true and and false per default, consistent with `LinearAlgebra`'s 5-arg `mul!`.
* add docstrings

### DIFF
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -27,7 +27,7 @@ Base.size(A::LinearMap, n) = (n==1 || n==2 ? size(A)[n] : error("LinearMap objec
 Base.length(A::LinearMap) = size(A)[1] * size(A)[2]
 
 Base.:(*)(A::LinearMap, x::AbstractVector) = mul!(similar(x, promote_type(eltype(A), eltype(x)), size(A, 1)), A, x)
-function LinearAlgebra.mul!(y::AbstractVector, A::LinearMap{T}, x::AbstractVector, α::Number=one(T), β::Number=zero(T)) where {T}
+function LinearAlgebra.mul!(y::AbstractVector, A::LinearMap, x::AbstractVector, α::Number=true, β::Number=false)
     length(y) == size(A, 1) || throw(DimensionMismatch("mul!"))
     if α == 1
         β == 0 && (A_mul_B!(y, A, x); return y)
@@ -52,7 +52,7 @@ function LinearAlgebra.mul!(y::AbstractVector, A::LinearMap{T}, x::AbstractVecto
     end
 end
 # the following is of interest in, e.g., subspace-iteration methods
-function LinearAlgebra.mul!(Y::AbstractMatrix, A::LinearMap{T}, X::AbstractMatrix, α::Number=one(T), β::Number=zero(T)) where {T}
+function LinearAlgebra.mul!(Y::AbstractMatrix, A::LinearMap, X::AbstractMatrix, α::Number=true, β::Number=false)
     (size(Y, 1) == size(A, 1) && size(X, 1) == size(A, 2) && size(Y, 2) == size(X, 2)) || throw(DimensionMismatch("mul!"))
     @inbounds @views for i = 1:size(X, 2)
         mul!(Y[:, i], A, X[:, i], α, β)
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -64,9 +64,9 @@ function LinearAlgebra.mul!(Y::AbstractMatrix, A::LinearMap, X::AbstractMatrix, 
     return Y
 end
 
-A_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector)  = mul!(y, A, x)
-At_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, transpose(A), x)
-Ac_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, adjoint(A), x)
+@inline A_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector)  = mul!(y, A, x)
+@inline At_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, transpose(A), x)
+@inline Ac_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, adjoint(A), x)
 
 # Matrix: create matrix representation of LinearMap
 function Base.Matrix(A::LinearMap)
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -5,6 +5,13 @@ export LinearMap
 using LinearAlgebra
 using SparseArrays
 
+if VERSION < v"1.2-"
+    import Base: has_offset_axes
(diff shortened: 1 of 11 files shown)


Jutho:
Lot's of activity here. I have been of the radar the last two weeks. I will try to start reviewing all of this next week; but I have other awaiting duties as well.

Regarding the `size` specialization; I think this code addition is not necessary. Surely the compiler will inline `size(linearmap, 1)` and eliminate the dead code associated with the creation of the tuple and computing its second element, only to retain the first element (and likewise when only the second element is required).

src/functionmap.jl
```diff
@@ -30,13 +30,31 @@ function Base.show(io::IO, A::FunctionMap{T}) where {T}
...
 Base.size(A::FunctionMap) = (A.M, A.N)
 LinearAlgebra.issymmetric(A::FunctionMap) = A._issymmetric
 LinearAlgebra.ishermitian(A::FunctionMap) = A._ishermitian
 LinearAlgebra.isposdef(A::FunctionMap)    = A._isposdef
 ismutating(A::FunctionMap) = A._ismutating
 _ismutating(f) = first(methods(f)).nargs == 3
 
+# special transposition behavior
+function LinearAlgebra.transpose(A::FunctionMap)
+    if A.fc!==nothing || A._issymmetric
```
Jutho: This should probably check that the linear map is real. We could also choose to make `transpose` and `adjoint` of a `FunctionMap` return a new function map, though this might not be type stable as it depends on `A._ishermitian`

Jutho:
Regarding the transpose/adjoint of a functionmap, I don't know what the check really buys. Suppose you did not define `fc`,maybe you still want to construct `B=A'`, only then to later only act with `B'`.

We could try to check that `ismutating` is compatible with both `f` and `fc` in the constructor of `FunctionMap` though.

dkarrasch:
Great to have you back and your eyes on it! The idea of this PR is to experiment with some potential changes and see whether they work at all, and finally cherry-pick whatever we feel is worth it. I don't have enough (maybe any) knowledge about what the compiler can eliminate and what not, but since this is not a great deal anyway, I'll remove the `size` stuff from this PR.

Making `adjoint`/`transpose` return a new `FunctionMap` is what I have experimented with in https://github.com/Jutho/LinearMaps.jl/pull/66/commits/44e7098565674689ada869aee0727bfde809d157, until I realized that in some cases we don't require `fc` to be defined and still compute the action of the adjoint/transpose. The good thing is that there haven't been any complaints about something not working, so leaving some things (for now) as they are, is not a bad solution.

src/functionmap.jl
dkarrasch: I'm not sure about the real check: we don't have one in the `At_mul_B!`, but use `f` for the transpose if `A` is symmetric in any case.

Regarding the new `FunctionMap` idea, as already mentioned I did experiment with that, but then I had two issues, IIRC. First, type stability as you say, and second I had difficulties of checking correctly for hermitianness of `A'A`, where `A::FunctionMap`.

src/LinearMaps.jl
```diff
@@ -57,9 +64,9 @@ function LinearAlgebra.mul!(Y::AbstractMatrix, A::LinearMap{T}, X::AbstractMatri
     return Y
 end
 
-A_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector)  = mul!(y, A, x)
-At_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, transpose(A), x)
-Ac_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, adjoint(A), x)
+@inline A_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector)  = mul!(y, A, x)
+@inline At_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, transpose(A), x)
+@inline Ac_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, adjoint(A), x)
```
Jutho: This too, I think, is not at all necessary. Such one-line definitions are automatically inlined by the compiler.

src/functionmap.jl
Jutho: Yes, but if the user specified that `A` is symmetric, then `f` can be used for both `A_mul_B` and `At_mul_B`, even if the action is complex-valued. But if it is not symmetric, then you only want to use `fc` for `At_mul_B` if the linear map is real, otherwise `fc` is supposed to be the adjoint, and thus different from the transpose.

Hence, the test (if it stays) should probably be:
`A._issymmetric || (eltype(A) <: Real && A.fc !== nothing)`

src/LinearMaps.jl
dkarrasch: Hm, pretty much all the one-line `mul!`s have an `@inline` macro in `LinearAlgebra`. When I was experimenting with the 5-arg mul!, I thought I had to add the macro somewhere to avoid a single allocation. But in that other PR, we have tests for that, so the situation might have been different. I think you're right that we shouldn't overload the code unless necessary for performance.

Jutho: That's strange, i.e. the comment about the `LinearAlgebra` statements. Well, I guess they do no harm so you can certainly leave them be.

dkarrasch: Yes, I just benchmarked `mul!(y, LinearMap(A), x)` vs. `mul!(y, A, x)` vs. `A_mul_B!(y, LinearMap(A), x)`, and there is no difference. I'll remove the macros then.

src/functionmap.jl
dkarrasch: You're right. I messed it up with all the combinations.

dkarrasch:
Perhaps you're right. The `FunctionMap`-related transposition/adjoint was working well, and the additional tests at construction don't buy anything. Eventually, the same error is thrown, and potentially users don't even run into the error when they transpose twice. I'll remove that part, and make a mental note that we may think about checking the in-place-ness of both functions, but keep that for another PR. That's not urgent.

dkarrasch:
Ready for review. 😉

Jutho:
Looks great; no comments; ready to merge.

PR #68, opened 2019-09-02
Use 5-arg mul! for matrices

### LINKED ISSUE #56: use LinearAlgebra's new generic in-place 5-arg mul! (requires Julia 1.3)?
see JuliaLang/julia#29634 - this may be relevant for the mul! definition at https://github.com/Jutho/LinearMaps.jl/blob/a024a4dd26cd4752e43142feda5cf3da8b5dc601/src/LinearMaps.jl#L23 and maybe some of the definitions at https://github.com/Jutho/LinearMaps.jl/blob/a024a4dd26cd4752e43142feda5cf3da8b5dc601/src/wrappedmap.jl#L27

ultimately I'm interested in using LinearMaps.jl with IterativeSolvers.jl in a way that avoids unnecessary allocations and dispatches to the fastest methods possible, for any linear maps that I construct.

dkarrasch:
Closes #56.

* introduce `MatrixMap{T} <: WrappedMap{T}`
* direct 5-arg mul! for MatrixMaps to `LinearAlgebra`'s 5-arg `mul!`
* make `MatrixMap` invariant under `adjoint`/`transpose`
* use this, if possible, in linear combinations

### DIFF
--- a/src/linearcombination.jl
+++ b/src/linearcombination.jl
@@ -82,6 +82,56 @@ function A_mul_B!(y::AbstractVector, A::LinearCombination, x::AbstractVector)
     end
     return y
 end
+function A_mul_B!(y::AbstractVector, A::LinearCombination{T,As}, x::AbstractVector) where {T, As<:Tuple{Vararg{MatrixMap}}}
+    # no size checking, will be done by individual maps
+    A_mul_B!(y, A.maps[1], x)
+    for n in 2:length(A.maps)
+        mul!(y, A.maps[n], x, true, true)
+    end
+    return y
+end
+
+function LinearAlgebra.mul!(y::AbstractVector, A::LinearCombination{T,As}, x::AbstractVector, α::Number=true, β::Number=false) where {T, As<:Tuple{Vararg{MatrixMap}}}
+    length(y) == size(A, 1) || throw(DimensionMismatch("mul!"))
+    if isone(α)
+        iszero(β) && (A_mul_B!(y, A, x); return y)
+        if isone(β)
+            for n in 1:length(A.maps)
+                mul!(y, A.maps[n], x, α, β)
+            end
+            return y
+        else # β != 0, 1
+            rmul!(y, β)
+            for n in 1:length(A.maps)
+                mul!(y, A.maps[n], x, α, β)
+            end
+            return y
+        end
+    elseif iszero(α)
+        iszero(β) && (fill!(y, zero(eltype(y))); return y)
+        isone(β) && return y
+        # β != 0, 1
+        rmul!(y, β)
+        return y
+    else # α != 0, 1
+        if iszero(β)
+            A_mul_B!(y, A, x)
+            rmul!(y, α)
+            return y
+        elseif isone(β)
+            for n in 1:length(A.maps)
+                mul!(y, A.maps[n], x, α, β)
+            end
+            return y
+        else # β != 0, 1
+            rmul!(y, β)
+            for n in 1:length(A.maps)
+                mul!(y, A.maps[n], x, α, true)
+            end
+            return y
+        end # β-cases
+    end # α-cases
+end
 
 end # VERSION
 
--- a/src/linearcombination.jl
+++ b/src/linearcombination.jl
@@ -46,6 +46,8 @@ LinearAlgebra.transpose(A::LinearCombination) = LinearCombination{eltype(A)}(map
 LinearAlgebra.adjoint(A::LinearCombination)   = LinearCombination{eltype(A)}(map(adjoint, A.maps))
 
 # multiplication with vectors
+if VERSION < v"1.3.0-alpha.115"
+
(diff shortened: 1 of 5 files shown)


dkarrasch:
I tested this on a `BlockMap`, in which subsequent matvec-mul results are added to previous results. The only allocations are due to views, in contrast to earlier Julia/LinearMaps versions.

dkarrasch:
Alright, this is tested really hard for zero allocations, for the case when the map is a wrapped map of some `AbstractMatrix` (called `MatrixMap`, not a type, just an abbreviation), a wrapped map of a uniformscaling object (called `UniformScalingMap`, which is going to be exported after #61 is merged), BTW these are joined as `FreeMap`s, and linear combinations of `FreeMap`s. In `BlockMap`s, any block can be the aforementioned `FreeMap` or LinearCombination of `FreeMap`s, and the only allocations are a few bytes for the views. Check out the newly added tests to get a flavor of what is allocation-free now.

@chriscoey Does this cover most of your applications? If not, I'd be interested to learn about the uncovered ones.

chriscoey:
@dkarrasch thank you, you're brilliant! I'll let you know in the next couple months as we run benchmarks of our as-yet-unreleased optimization solver that calls methods from IterativeSolvers

chriscoey:
Is this one ready for merge?

dkarrasch:
I think it is, but I would love to have @Jutho's eyes on it, just in case.

Jutho:
My apologies for the delay, I'll try to review before the end of the week.

src/linearcombination.jl
```diff
@@ -60,6 +62,67 @@ function A_mul_B!(y::AbstractVector, A::LinearCombination, x::AbstractVector)
...
+    # no size checking, will be done by individual maps
+    A_mul_B!(y, A.maps[1], x)
+    l = length(A.maps)
+    if l>1
+        z = similar(y)
+        for n in 2:l
+            An = A.maps[n]
+            if An isa FreeMap
+                mul!(y, An, x, true, true)
+            else
```
Jutho: One trick I learned recently is to add, in the body of a loop, a statement like
```julia
if !(@isdefined z)
     z = similar(y)
end
```
which would then only allocate `z` once the first map which is not a `FreeMap` is encountered, and thus not allocate at all if all the maps are `FreeMap`

src/linearcombination.jl
```diff
@@ -60,6 +62,67 @@ function A_mul_B!(y::AbstractVector, A::LinearCombination, x::AbstractVector)
...
+                A_mul_B!(z, A.maps[n], x)
+                y .+= z
+            end
+        end
+    end
+    return y
+end
+function A_mul_B!(y::AbstractVector, A::LinearCombination{T,As}, x::AbstractVector) where {T, As<:Tuple{Vararg{FreeMap}}}
```
Jutho: I did not see this yet, but it would then become unnecessary with the trick above.

src/wrappedmap.jl
```diff
@@ -17,6 +17,22 @@ function WrappedMap{T}(lmap::Union{AbstractMatrix, LinearMap};
...
 end
 
+const MatrixMap{T} = WrappedMap{T,<:AbstractMatrix}
+
+LinearAlgebra.transpose(A::MatrixMap{T}) where {T} =
+    WrappedMap{T}(transpose(A.lmap); issymmetric=A._issymmetric, ishermitian=A._ishermitian, isposdef=A._isposdef)
+LinearAlgebra.adjoint(A::MatrixMap{T}) where {T} =
+    WrappedMap{T}(adjoint(A.lmap); issymmetric=A._issymmetric, ishermitian=A._ishermitian, isposdef=A._isposdef)
+
+Base.:(==)(A::MatrixMap, B::MatrixMap) = (eltype(A) == eltype(B) && A.lmap == B.lmap)
```
Jutho: I would be inclined to also check the specifiers `_issymmetric` etc here. If a user went through the effort of creating one `WrappedMap` where he set `issymmetric` to true, and another one with the same matrix where he set `issymmetric` to false, he probably wants them to be different. (Not that I see this happening in any realistic use case).

src/linearcombination.jl
Jutho: On the other hand, the current implementation is also very Julian, so just keep it like it is. Maybe just change the order of these two methods, such that one first reads the implementation of the more specific method and then the one of the more general fallback method.

dkarrasch:
Thanks, @Jutho! Things are a bit spread over the place, but the idea is to have `true` and `false` as defaults for alpha and beta everywhere. As for the other `mul!` methods, this is done in #66 (I hope, but will make sure).

Jutho:
Yes indeed; I should have remembered. Anyway, I think this is good to merge, aside from maybe the `==` method for `MatrixMap`. And please remind me of any open PRs which are ready for review (and feel free to keep reminding me if there is no reaction), because I have not followed closely.

PR #69, opened 2019-09-12
handle uniform scalings in compositions of maps explicitly

dkarrasch:
This is probably an incredibly dirty hack, but it does all the things we were considering in the context of #57, namely:
* it's type stable, since it doesn't touch the constructors at all, but rather handles everything within the `mul!` as suggested by @Jutho in #63;
* it finishes earlier when one of the maps in a composition is a scaling by zero;
* it skips uniform scalings by `one`, no matter in which position they occur, and also no matter of what exact type the `one` is, i.e., this is not restricted to `Bool`s.

### DIFF
--- a/src/composition.jl
+++ b/src/composition.jl
@@ -108,25 +108,67 @@ function A_mul_B!(y::AbstractVector, A::CompositeMap, x::AbstractVector)
         A_mul_B!(y, A.maps[1], x)
     else
         T = eltype(y)
-        dest = Array{T}(undef, size(A.maps[1], 1))
-        A_mul_B!(dest, A.maps[1], x)
-        source = dest
-        if N>2
-            dest = Array{T}(undef, size(A.maps[2], 1))
-        end
-        for n in 2:N-1
-            try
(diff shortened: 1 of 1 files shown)


src/composition.jl
```diff
@@ -108,25 +108,67 @@ function A_mul_B!(y::AbstractVector, A::CompositeMap, x::AbstractVector)
         A_mul_B!(y, A.maps[1], x)
     else
         T = eltype(y)
-        dest = Array{T}(undef, size(A.maps[1], 1))
```
Jutho: The reason for assigning `dest` this way is to make sure that the `eltype` of `dest` is fixed (to that of `y`) such that it can hopefully be reused later on if there are more than 2 linear maps.  This might fail with the new lines of code below.

Jutho:
Thanks for working on this. 

While the tests turn green, I think I could see cases where this fails. Also, there is still copying going on with identity linear maps, so I think a more involved change is required if we really don't want to do any extra work for uniformscalingmap with lambda = 1.

PR #70, opened 2019-09-19
Return correct size of BlockMap involving zero-size blocks

perrutquist:
This fixes a bug where trying to get the size of a BlockMap would
fail if the last block had size zero in the relevant dimension.

### DIFF
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -48,7 +48,7 @@ function rowcolranges(maps, rows)::Tuple{Vector{UnitRange{Int}},Vector{UnitRange
 end
 
 function Base.size(A::BlockMap)
-    return A.rowranges[end][end], A.colranges[end][end]
+    return last(A.rowranges[end]), last(A.colranges[end])
 end
 
 ############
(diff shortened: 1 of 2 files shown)


dkarrasch:
Out of curiosity, why would you have zero-size blocks? When writing that code, I did not have zero-size blocks in mind _at all_. So I'm pleased to see that the colrange-rowrange determination made some sense (up to the bug you fixed) even in that case.

src/blockmap.jl
```diff
@@ -48,7 +48,7 @@ function rowcolranges(maps, rows)::Tuple{Vector{UnitRange{Int}},Vector{UnitRange
 end
 
 function Base.size(A::BlockMap)
-    return A.rowranges[end][end], A.colranges[end][end]
+    return last(A.rowranges[end]), last(A.colranges[end])
```
dkarrasch: Could you please turn this into a one-liner?
```julia
Base.size(A::BlockMap) = (last(A.rowranges[end]), last(A.colranges[end]))
```
This is the format that I have in #66 for this function anyway.

chriscoey:
Speaking for myself, I sometimes have zero-size blocks in the KKT conditions of optimization problems, for example in case there are no primal equality constraints

dkarrasch:
@perrutquist Thank you very much!

PR #71, opened 2019-09-23
RFC: allow matrices in BlockMaps

dkarrasch:
This is a slightly controversial follow-up on #52, to allow `AbstractMatrix` in concatenation. The change is straightforward, but it increases compilation time significantly. Even after it's precompiled, loading it takes seemingly forever:
```julia
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.3.0-rc2.0 (2019-09-12)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia> @time using LinearMaps
 59.447319 seconds (277.36 M allocations: 16.037 GiB, 2.30% gc time)
```
vs. current master
```julia
julia> @time using LinearMaps
  0.144496 seconds (219.51 k allocations: 13.011 MiB)
```
Maybe I did something wrong, or maybe it's not really worth it after all...


Jutho:
I don't see anything wrong. That's quite disturbing indeed. How does the maximum number of arguments (i.e. currently 8) in our `@eval` section affects the compilation time?

dkarrasch:
I'll try to drastically decrease that number, and see what happens.

dkarrasch:
It affects it dramatically. Here are some `@time using LinearMaps` results after precompilation:
* `k in 1:7`: 25.101010 seconds (134.24 M allocations: 7.760 GiB)
* `k in 1:6`: 3.497141 seconds (18.72 M allocations: 1.081 GiB)
* `k in 1:5`: 0.734655 seconds (2.66 M allocations: 157.440 MiB)
Honestly, I don't know what is going on in the background, and I have no feeling about what minimum number would be required for this feature to be useful. I already "got used" to 8. 😃

dkarrasch:
Also, something is going incredibly wrong on Linux CI...

Would be great to identify the root cause for the slowdown and try to get a resolution at the Julia level.

chriscoey:
Anyone we can ping to figure out what's going on?

dkarrasch:
After some minor fixes, this does actually work now locally, so you can do
```julia
[LinearMap(A) A A]
```
with some matrix `A`. The compile time is still massive. I believe this could be due to these method invalidations that have become a hot topic on Julia master recently. I don't understand how to diagnose method invalidations or how to avoid them (yet), but I hope some insights of the masterminds trickle down eventually, because having this feature would be really nice and useful.

src/blockmap.jl
```diff
@@ -92,7 +92,7 @@ julia> L * ones(Int, 6)
 """
 Base.hcat
 
-function _hcat(As::Union{LinearMap,UniformScaling}...)
+function Base.hcat(As::Union{LinearMap,UniformScaling,AbstractVecOrMat}...)
```
dkarrasch, line 95: The AbstractVectors are here to avoid a method ambiguity with LinearAlgebra. This may be slightly controversial, because we don't ever interpret vectors as column matrices, but the use case here seems fairly specific, and why not interpret a vector in [h/v/hv]cat as a column of a bigger "matrix"?

Also, we haven't done this direct Base method overload originally because we were afraid that we may re-define Base behavior. But my understanding is that since our Union here is strictly bigger than what we have in Base/LinearAlgebra, our method is more generic, and hence does not get called when the arguments are only matrices and uniform scalings. I have added detailed tests for which method gets called when used without LinearMaps, and moreover, I am not getting any ambiguity errors, so I believe this should be fine.

And finally, this allows us to drop the requirement of having a LinearMap among the first 8 arguments. It seems like everything is just handled by dispatch.

src/blockmap.jl
```diff
@@ -445,17 +447,12 @@ for k in 1:8 # is 8 sufficient?
     mapargs = ntuple(n -> :(LinearMap($(Symbol(:A,n)))), Val(k-1))
     # yields (:LinearMap(A1), :LinearMap(A2), ..., :LinearMap(A(k-1)))
 
-    @eval begin
-        function SparseArrays.blockdiag($(Is...), $L, As::Union{LinearMap,AbstractMatrix}...)
+    # since the below method is more specific than the Base method, it would redefine Base behavior
+    @eval function Base.cat($(Is...), $L, As::MapOrMatrix...; dims::Dims{2})
```
dkarrasch, line 451: The rationale above does not work here: `cat` is defined for generic arguments, so our method would be more specific and hence take over matrix concatenation, for instance. That's why we need to keep this requirement that among the first 8 arguments there is a `LinearMap`.

dkarrasch: Hm, I just noticed that it doesn't apply to `blockdiag` either. We'd be overloading `SparseArrays.blockdiag` for `AbstractMatrix`s, i.e., not our own type. 🏴‍☠️

dkarrasch:
I'm confident now that the direct overload approach should be acceptable. First, we have methods in `Base` like

https://github.com/JuliaLang/julia/blob/7af47af9c2b478248c650c5d3785e3eaa0f0e6fc/base/abstractarray.jl#L1458-L1459

and then

https://github.com/JuliaLang/julia/blob/7af47af9c2b478248c650c5d3785e3eaa0f0e6fc/stdlib/LinearAlgebra/src/uniformscaling.jl#L383-L404

which are more generic (and hence don't override the previous `Base` behavior) because of the `Union`. Now, if we have
```julia
Base.hcat(As::Union{AbstractVecOrMat,LinearMap,UniformScaling}...)
```
then ours is even more generic and, by the same rule, does not override the previous `Base`/`LinearAlgebra` behavior if only matrices and uniform scalings are passed as arguments. The same argument does not apply to `cat(A...; dims=(1,2))` (and  one might argue for `SparseArrays.blockdiag` as well), because there is only the generic

https://github.com/JuliaLang/julia/blob/7af47af9c2b478248c650c5d3785e3eaa0f0e6fc/base/abstractarray.jl#L1725

so ours would be more specific and override pure `AbstractMatrix` handling, for instance. With `blockdiag`, it's slightly different. Here, we have the specific methods

https://github.com/JuliaLang/julia/blob/7af47af9c2b478248c650c5d3785e3eaa0f0e6fc/stdlib/SparseArrays/src/sparsematrix.jl#L3334-L3344

so if we defined a `SparseArrays.blockdiag(A::MapOrMatrix...)` method, it would allow to blockdiagonally concatenate general matrices, and return a `BlockDiagonalMap`. Actually, I don't think this is absurd, as we would provide a lazy representation of the block diagonal operator...

dkarrasch:
@Jutho What's your schedule these days? Will you have time for a review of individual PRs, or should I merge and leave the review for your overhaul? I would say this is ready, and would be a fantastic feature for v3.0.

Jutho:
I suggest that you merge this if you feel satisfied with it. I have not been much involved with the whole `BlockMap` implementation, so it makes more sense that I review the whole code at once rather than this specific PR. I anyway won't have much time this week, but I will try to make time for a proper review next week.

dkarrasch:
Cool, I'll merge all my 20 PRs then and go get a drink. 😂 🍸

PR #72, opened 2019-10-16
BlockMap enhancements

dkarrasch:
* add 5-arg `mul!` for `BlockMap`s
* use `Base.@propagate_inbounds` in `BlockMap`-multiplication
* add overhead-free `BlockDiagonalMap`
* add 5-arg `mul!(::AbstractMatrix, ::BlockMap, ::AbstractMatrix, alpha, beta)`

PR #74, opened 2019-11-18
WIP: get rid of A*_mul_B!

dkarrasch:
This started innocently, and became a major code overhaul. I realized somewhere in #72 I guess that one can get rid of all the `A*_mul_B!` by dispatching`mul!` on things like `AdjointMap{<:Any,<:SomeLinearMap}`. This turned out to be really nice, because for subtypes of `LinearMap`s, that are invariant under transposition/adjoint, certain combinations are now impossible to obtain by standard means, even when artificially wrapping, say, a `CompositeMap` in a `LinearMap` and only then taking the tranpose/adjoint. The beauty of it is that, eventually, up to some helper functions, this package is back where at was pre-0.7, that is, provides types and function overloads for `Base` or `LinearAlgebra` functions (with `kronsum` and some symbols the only exceptions).

For review, I suggest to skip `blockmap.jl`, because that may require a rebase after #72 is merged (even though I simplified some code already).

EDIT: I forgot to mention that this also takes big steps forward in providing specialized 5-arg `mul!`s for both vectors **and** matrices.

### DIFF
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -109,10 +109,6 @@ Base.@propagate_inbounds function LinearAlgebra.mul!(Y::AbstractMatrix, A::Linea
     return Y
 end
 
-# A_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector)  = mul!(y, A, x)
-# At_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, transpose(A), x)
-# Ac_mul_B!(y::AbstractVector, A::AbstractMatrix, x::AbstractVector) = mul!(y, adjoint(A), x)
-
 include("transpose.jl") # transposing linear maps
 include("wrappedmap.jl") # wrap a matrix of linear map in a new type, thereby allowing to alter its properties
 include("uniformscalingmap.jl") # the uniform scaling map, to be able to make linear combinations of LinearMap objects and multiples of I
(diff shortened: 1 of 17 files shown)


dkarrasch:
Currently, tests are failing in the linearcombinations tests, where a single allocation counting test fails. I can't reproduce this locally (on macos), tests pass all fine here on all available Julia versions. Would be great if somebody could check out this branch and see if that is spurious. I did notice though that 5-arg mul! with matrices is certainly not allocation-free on Julia v1.3.rc's.
```julia
using LinearAlgebra, BenchmarkTools
A = rand(100,200)
x = rand(size(A, 2), 10)
y = rand(size(A, 1), 10)
α = rand()
β = rand()
@benchmark mul!($y, $A, $x, $α, $β)

BenchmarkTools.Trial: 
  memory estimate:  64 bytes
  allocs estimate:  3
  --------------
  minimum time:     19.190 μs (0.00% GC)
  median time:      20.584 μs (0.00% GC)
  mean time:        24.629 μs (0.00% GC)
  maximum time:     475.044 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1
```
And it does seem to cost performance. I assume this is due to the generation of the `MulAdd` structure. At least, on Julia v1.4, these are gone (again, locally), and the above median time is significantly lower!

dkarrasch:
This PR requires a little work to merge the other PRs in, e.g., the mulstyle trait. When those are merged, I'll finalize this one and shout again for review.

JeffFessler:
I hope to return to #65 (`IndexMap`) after this simplification process is finalized.

dkarrasch:
This is totally outdated. I took some of the benchmarks and put them in #116. The final thing that this PR tried to address is to allocate only a single intermediate vector when doing 5-arg map-matrix multiplication for a `ThreeArg` mulstyle. But that requires a fresh approach, perhaps.

PR #78, opened 2019-12-16
Minor updates in README.md

wsshin:
- Fix a typo: "spefified" → "specified"
- In the explanation of the `LinearMap` constructor taking a function, mention the argument name `fc` in the explanation of the second optional argument for clarity.


dkarrasch:
Many thanks @wsshin!

PR #80, opened 2019-12-27
select intermediate destination of CompositionMap based on input

kose-y:
I was trying to use `LinearMap` with a CuArray, and this issue blocked me.

### DIFF
--- a/src/composition.jl
+++ b/src/composition.jl
@@ -126,12 +126,11 @@ function A_mul_B!(y::AbstractVector, A::CompositeMap, x::AbstractVector)
     if N==1
         A_mul_B!(y, A.maps[1], x)
     else
-        T = eltype(y)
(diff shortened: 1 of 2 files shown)


dkarrasch:
Thanks, this LGTM. While you're at it, could you please also replace the two `Array` constructors in `kronecker.jl`, lines 109-110?

kose-y:
Sure.

dkarrasch:
Thanks once more. Now I realize that we may actually not need the `T = eltype(y)`/`Ty = eltype(y)` stuff (in both cases), since the default is `y`'s eltype anyway. Could you please remove them?

dkarrasch:
Many thanks for your contribution. I'll release this soonish after #72 is merged as v2.6.

Jutho:
Much better indeed, thanks.

PR #82, opened 2020-01-08
Keep the FunctionMap promise

dkarrasch:
In our README, we promise that either(!) of the `FunctionMap` properties `issymmetric` and `ishermitian` allows to compute both `A'*x` and `transpose(A)*x`. Currently, this is not the case, so I've added this functionality.


dkarrasch:
These are new features, that would have errored before. They don't interfere with previous multiplication routines, since the new branches are inserted after what we had before and before the error throwing part.

PR #83, opened 2020-01-28
allow direct multiplication with out-of-place transpose/adjoint FunctionMaps

dkarrasch:
xref: https://discourse.julialang.org/t/linearmaps-transpose-function-not-working-the-same-as-forward-function/33759

### DIFF
--- a/src/functionmap.jl
+++ b/src/functionmap.jl
@@ -48,6 +48,64 @@ function Base.:(*)(A::FunctionMap, x::AbstractVector)
     end
     return y
 end
+function Base.:(*)(A::AdjointMap{<:Any,<:FunctionMap}, x::AbstractVector)
+    Afun = A.lmap
+    ishermitian(Afun) && return Afun*x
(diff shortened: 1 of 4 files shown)


src/functionmap.jl
```diff
@@ -78,7 +136,7 @@ function At_mul_B!(y::AbstractVector, A::FunctionMap, x::AbstractVector)
 end
 
 function Ac_mul_B!(y::AbstractVector, A::FunctionMap, x::AbstractVector)
-    (ishermitian(A) || (isreal(A) && issymmetric(A))) && return A_mul_B!(y, A, x)
```
dkarrasch: There is no way to construct an `AdjointMap{<:Real,<:FunctionMap}` anyway.

dkarrasch:
Maybe we should add tests with `FunctionMap`s à la `A = Symmetric(rand(ComplexF64, 10, 10))` and `LinearMap(x -> A*x, 10, 10)`...

Jutho:
Looks good!

PR #89, opened 2020-04-03
Move CI to Github action

dkarrasch:


Jutho:
I have not been following the move of CI to github actions; I am certainly not opposed to this, but was just wondering (for my other packages): is their a clear advantage to do this sooner better than later? And how much compute time do you get as a free user? Are there limits/clear specifications?

dkarrasch:
The main distinction is whether packages are public or private, if I recall correctly. I was annoyed (also with my other package) that Travis wasn't showing up in the PR webpage anymore, so not showing success or failure. Maybe one could have fixed that, but my impression is that you get computing power much faster with Github than with Travis. I think I also read this somewhere else. At times, I was waiting for 20 minutes or more on Travis just to get something started. With Github, it starts right away. There is some restriction on Mac testing on Github, IIRC, but for this package, where tests run at most 10 minutes, we are freeing the machines quickly, and I believe we're flying under the radar.

PR #92, opened 2020-06-16
move CompositeMap multiplication code to helpers

### LINKED ISSUE #90: Avoiding repeated allocations in mul! for CompositeMap
Replacing `y=A*x` with `mul!(y, A, x)` gives the (perhaps naive) user (e.g., me) the sense that no memory allocations will need to be made, and often that is the case.
But if `A` is a `CompositeMap`, e.g., `A = B * C`, then internally `A*x = B*(C*x)` and the current implementation will allocate a temporary variable for the product `C*x` every time we do `A*x` or the `mul!` version.  This overhead adds up when iterating.

We know the size of `C` at the time we make the `CompositeMap`, so in principle the LM package could allocate a work space for it and store it as part of the struct, so that `mul!` operations would need no new allocations.  (We'd also need a work space for `B'` for the adjoint.)
Perhaps this could be a user-selectable option when creating a `LinearMap`.

Thoughts on this feature suggestion?  I might try it myself unless anyone sees flaws in it...

dkarrasch:
One way to close #90. @JeffFessler 

Actually, independent of #90 I like this a lot, because it moves all those case distinctions to the type domain.

### DIFF
--- a/src/composition.jl
+++ b/src/composition.jl
@@ -124,19 +124,19 @@ function A_mul_B!(y::AbstractVector, A::CompositeMap{T,<:Tuple{LinearMap}}, x::A
     return A_mul_B!(y, A.maps[1], x)
 end
 function A_mul_B!(y::AbstractVector, A::CompositeMap{T,<:Tuple{LinearMap,LinearMap}}, x::AbstractVector) where {T}
-    compositemul!(y, A, x, similar(y, size(A.maps[1], 1)))
+    _compositemul!(y, A, x, similar(y, size(A.maps[1], 1)))
 end
 function A_mul_B!(y::AbstractVector, A::CompositeMap{T,<:Tuple{Vararg{LinearMap}}}, x::AbstractVector) where {T}
-    compositemul!(y, A, x, similar(y, size(A.maps[1], 1)), similar(y, size(A.maps[2], 1)))
+    _compositemul!(y, A, x, similar(y, size(A.maps[1], 1)), similar(y, size(A.maps[2], 1)))
 end
 
-function compositemul!(y::AbstractVector, A::CompositeMap{T,<:Tuple{LinearMap,LinearMap}}, x::AbstractVector, z::AbstractVector) where {T}
+function _compositemul!(y::AbstractVector, A::CompositeMap{T,<:Tuple{LinearMap,LinearMap}}, x::AbstractVector, z::AbstractVector) where {T}
     # no size checking, will be done by individual maps
     A_mul_B!(z, A.maps[1], x)
     A_mul_B!(y, A.maps[2], z)
     return y
 end
-function compositemul!(y::AbstractVector, A::CompositeMap{T,<:Tuple{Vararg{LinearMap}}}, x::AbstractVector, source::AbstractVector, dest::AbstractVector) where {T}
+function _compositemul!(y::AbstractVector, A::CompositeMap{T,<:Tuple{Vararg{LinearMap}}}, x::AbstractVector, source::AbstractVector, dest::AbstractVector) where {T}
     # no size checking, will be done by individual maps
     N = length(A.maps)
     A_mul_B!(source, A.maps[1], x)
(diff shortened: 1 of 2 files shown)


JeffFessler:
I agree that this should help simplify working on #90.  (If you were saying it will close it then I don't understand because it still has `similar` as an allocator in it.)

I notice the code coverage did not cover the case `A::CompositeMap{T,Tuple{LinearMap}}`
I actually don't know how that case would ever arise because I thought we needed more than one map to make a `CompositeMap`?

dkarrasch:
> I agree that this should help simplify working on #90. (If you were saying it will close it then I don't understand because it still has `similar` as an allocator in it.)

Well, you can skip the `mul!` calls and call `compositemul!` (or whatever it gets called at the end) instead, when you know what you're doing. Or, you can write another `LinearMap` subtype (along the lines of `CompositeMap`) that has temp variable fields and, in its own `mul!`, calls `compositemul!` and passes its memory fields as fourth and fifth arguments.
 
> I notice the code coverage did not cover the case `A::CompositeMap{T,Tuple{LinearMap}}`
> I actually don't know how that case would ever arise because I thought we needed more than one map to make a `CompositeMap`?

Technically, you could construct one manually:
```julia
LinearMaps.CompositeMap{Float64}((LinearMap(rand(3,3)),))
```
We had three lines of code for handling that theoretical case already, so this is not new, and missing coverage was perhaps not detected earlier. I'll add a quick test to cover that case.

dkarrasch:
Since this is only code rearrangement, I guess this by itself is not controversial, but I'll leave it open for a few days before merging.

JeffFessler:
I've come to appreciate the potential elegance of this `compositemul!` approach and I started to implement something that will use it.  But I wanted to be general so I started with the 5-arg version of `mul!`.  Should there be a corresponding 5-arg version of `compositemul!` (that will have 7 args by the time you add alpha and beta)?

I'll readily admit I'm not clear on how the 5-arg version is even working.  I guess it is handled by ` LinearAlgebra.mul!` but that routine has the line `y .+= rmul!(A * x, α)`.  A small test with `@allocated` suggests that line does an allocation (presumably for the intermediate product `A * x`) so it seems not to be a fully in-place operation in the way I hoped.
(Yet `@allocated mul!(y, A, x, 5, 2)` suggests that no memory is allocated which baffles me.)
If the 3-arg `mul!` suffices that would be great... Thoughts?

```
using LinearAlgebra: mul!, rmul!
using Test: @allocated

A = randn(2^9,8)
x = ones(8)
y = zeros(2^9)
#@allocated mul!(y, A, x, 5, 2) # 0!?
@allocated y .+= rmul!(A*x, 5) # 4272 bytes
```

dkarrasch:
> Line 119 should say `CompositeMap`

Copy, paste, 🤦

JeffFessler:
I suspect that `UniformScalingMap` could be handled more efficiently (memory wise).
If `B = 5A` then `mul!(y, B, x)` can be done without allocation via `mul!(y, mul!(y, A, x), 5)`.

dkarrasch:
Yes. I think @Jutho suggested earlier to introduce a "`ScaledMap`". Your example should actually be implemented by `mul!(y, A, x, 5)`. In case `A` is matrix-based, this could be even computed in one pass over `y`.

JeffFessler:
Oh gosh, I hadn't noticed there was a 4-arg `mul!` in there.  I might be willing to help with a `ScaledMap` if you agree that it is useful, but I think I might wait for #74 in hopes that it simplifies things...  Thanks.

dkarrasch:
>  I might be willing to help with a ScaledMap if you agree that it is useful,

Be our guest 😛 . I don't think #74 is worth pursuing. I should actually close it, open a PR with extensive comments instead, but leave the branch for orientation. It contains some nice details, but the overall approach is counterproductive the effort of allowing to pass intermediate storage, I think. Shortly, the reason is that if we don't have any multiplication functions other than `mul!` overloads, then how are we going to pass vectors for intermediate results, for instance? I may be wrong, so I'll need to lay out my thoughts in some detail, so others can agree or disagree. But I believe this is a much larger project, that can be put on top of a `ScaledMap`. Notice, however, that using 4- or 5-arg `mul!` is only appropriate for `eltype`s `<:Union{Real,Complex}`, because the scalar is supposed to be applied from the left in 4-arg `mul!`.

PR #93, opened 2020-06-17
remove obsolete comparison functions

dkarrasch:
IIUC, these seem to do nothing but the generic fall back.

### DIFF
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -288,12 +288,8 @@ function LinearAlgebra.ishermitian(A::BlockMap)
 end
 
 ############
-# comparison of BlockMap objects, sufficient but not necessary
-############
-
-Base.:(==)(A::BlockMap, B::BlockMap) = (eltype(A) == eltype(B) && A.maps == B.maps && A.rows == B.rows)
-
 # special transposition behavior
+############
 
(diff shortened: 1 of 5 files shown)


Jutho:
Are you sure? The docs say that `==` falls back to `===` for generic types, and then the docs for `===` state

> Determine whether x and y are identical, in the sense that no program could distinguish them. First the types of x and y are compared. If those are identical, mutable objects are compared by address in memory and immutable objects (such as numbers) are compared by contents at the bit level. This function is sometimes called "egal". It always returns a Bool value.

In principle, the `LinearMap` subtypes are immutable, but they are not `isbits`, since they will contain pointers to underlying matrices or functions. You can try to see if (before this PR), `==` and `===` give the same answer, if you do `e.g.`

```julia
A = randn(3,3)
B = LinearMap(A) + LinearMap(A)'
C = LinearMap(copy(A)) + LinearMap(copy(A))'
B == C # true
B === C # false
```
So I think that after this PR, `B == C` will also return `false`. But maybe this is a convoluted example.

dkarrasch:
Right, I was struggling to construct an example. Thanks, @Jutho, I'll close this.

PR #96, opened 2020-07-17
Remove defaults for alpha and beta in 5-arg mul!

dkarrasch:
I think this was not quite correct in the first place, and it was likely the cause for all my trouble in #74. In `LinearAlgebra`, there is also no defaults for alpha and beta in 5-arg `mul!`, but the 3-arg `mul!` further calls the 5-arg `mul!` with `true` and `false`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ jobs:
   test:
     name: Julia ${{ matrix.version }} - ${{ matrix.os }}
     runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.version == 'nightly' }}
     strategy:
       matrix:
         version:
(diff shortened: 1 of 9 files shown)


dkarrasch:
I included a little `BlockMap` optimization. As we go along a row, the multiplication process is somewhat similar to a `LinearCombination`-`mul!`. Since we have the trait `MulStyle` and some infrastructure in `LinearCombination`s to allocate an intermediate vector at most once (and not at all, if possible), this extends the same approach to `BlockMap`s. Conversely, for very specialized code, one can hook into the helper functions and provide intermediate storage from the outside, as possible with `CompositeMap`s (#92). I think this is ready for review, if anyone wants to; since this is rather minor and very well tested, I'd merge and release soon.

PR #87, opened 2020-03-28
Introduce specific show methods

dkarrasch:
I was very inspired by the `show` methods in #84, so I went ahead and introduced consistent such methods for all `LinearMap` subtypes. This is the current output:
```julia
julia> using LinearAlgebra, LinearMaps

julia> A = LinearMap(rand(10, 10))
10×10 LinearMaps.WrappedMap{Float64} of
 0.6769273787529222   0.036539631768852665  …  0.9873418166811534
 0.6765549649256377   0.9624717119719399       0.3575343347655404
 0.31545830565381205  0.846611890984601        0.07509443800525162
 0.9046438881682486   0.6282119777830519       0.16271650036125607
 0.13027892373965533  0.18887153454128924      0.3656173314780382
 0.8105579896722099   0.8689278012808201    …  0.613575621252296
 0.3501936207693894   0.41662154485059477      0.634150209365973
 0.5371570288242542   0.7182968669111338       0.3200262911743803
 0.9126028660250189   0.25564178492353906      0.23743297526797136
 0.5819814278543964   0.9943728715581492       0.4875339524077975

julia> B = LinearMap(rand(10, 10))
10×10 LinearMaps.WrappedMap{Float64} of
 0.8700744216295719     0.458876698914481    …  0.21599839571588952
 0.5444272069489469     0.30974679189517595     0.20158982970939365
 0.0039023841368437395  0.1743141153244383      0.5255090943219167
 0.6585137742811065     0.21529067784534073     0.09553647923800157
 0.6810489351017359     0.7726713554570053      0.88655004487403
 0.02948775918770319    0.804402117644776    …  0.8520117918594712
 0.981317366746389      0.42937017270368405     0.723298669979153
 0.05590009791874673    0.23243683344460941     0.7504218236329141
 0.6105003794996771     0.23372276641277856     0.5026331262236272
 0.45419955654642186    0.9829667847054768      0.5560613057463222

julia> A + B
10×10 LinearMaps.LinearCombination{Float64} with 2 maps:
10×10 LinearMaps.WrappedMap{Float64} of
 0.6769273787529222   0.036539631768852665  …  0.9873418166811534
 0.6765549649256377   0.9624717119719399       0.3575343347655404
 0.31545830565381205  0.846611890984601        0.07509443800525162
 0.9046438881682486   0.6282119777830519       0.16271650036125607
 0.13027892373965533  0.18887153454128924      0.3656173314780382
 0.8105579896722099   0.8689278012808201    …  0.613575621252296
[...]

PR #100, opened 2020-07-29
Left vector multiply

### LINKED ISSUE #99: Left multiply with Adjoint vector
There is currently a slight inconsistency that arises when left multiplying by a `Transpose` or `Adjoint` vector due to the current design where `y' *A` returns a `LinearMap` because `y'` is treated as a `AbstractArray` rather than as the more specific `Adjoint` vector.  Specifically, the following example shows that there is a slight departure of associativity:
 
```
using LinearMaps: LinearMap
A = LinearMap(ones(3,4))
y = ones(3)
x = ones(4)
y' * (A * x) # scalar
(y' * A) * x # 1-element array
```
Given that 'A*x' returns a vector, I would propose that `y'*A` should return an `Adjoint` vector rather than a `LinearMap` (when `y'` is an `Adjoint` vector).  That (slightly breaking change) would "restore" the associativity.   If `Y'` is an `Adjoint Matrix` then the current behavior seems fine and is consistent with how `A*X` works for a `LinearMap` multiplied by a `Matrix`.  It is only the adjoint vector case that needs refinement.

This issue arose when I was refining the tests for left multiplication for the "multidimensional" versions of
https://github.com/JeffFessler/LinearMapsAA.jl.  Obviously it is not urgent since no one reported it previously, but I would be willing to make a PR if there is agreement that `y'*A` should return an `Adjoint` vector.

Edit: I think it is easy to address for ordinary `*` but I might need help with the `mul!` version.  Or perhaps it is not so important to support `mul!` for this case.

JeffFessler:
Closes #99 by providing `*` and `mul!` methods for adjoint/transpose vectors on left of a `LinearMap`, like `y'*A`.

### DIFF
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -52,12 +52,10 @@ function check_dim_mul(C, A, B)
     # @info "checked vector dimensions" # uncomment for testing
     mA, nA = size(A) # A always has two dimensions
     mB, nB = size(B, 1), size(B, 2)
-    if mB != nA
-        throw(DimensionMismatch("left factor has dimensions ($mA,$nA), right factor has dimensions ($mB,$nB)"))
-    end
-    if size(C, 1) != mA || size(C, 2) != nB
+    (mB == nA) ||
+        throw(DimensionMismatch("left factor has dimensions ($mA,$nA), right factor has dimensions ($mB,$nB)")
+    (size(C, 1) != mA || size(C, 2) != nB) &&
         throw(DimensionMismatch("result has dimensions $(size(C)), needs ($mA,$nB)"))
-    end
     return nothing
 end
 
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -47,20 +47,17 @@ Base.ndims(::LinearMap) = 2
 Base.size(A::LinearMap, n) = (n==1 || n==2 ? size(A)[n] : error("LinearMap objects have only 2 dimensions"))
 Base.length(A::LinearMap) = size(A)[1] * size(A)[2]
 
-# check dimension consistency for right multiply: y = A*x and Y = A*X
-function check_dim_mul(Y::AbstractVecOrMat, A::LinearMap, X::AbstractVecOrMat)
+# check dimension consistency for multiplication C = A*B
+function check_dim_mul(C, A, B)
     # @info "checked vector dimensions" # uncomment for testing
-    m, n = size(A)
-    (m == size(Y, 1) && n == size(X, 1) && size(Y, 2) == size(X, 2)) ||
-        throw(DimensionMismatch("mul! $(size(X)) $(size(Y)) $(size(A))"))
-    return nothing
-end
-
-# check dimension consistency for left multiply: X = Y*A (e.g., Y=y')
-function check_dim_mul(X::AbstractVecOrMat, Y::AbstractVecOrMat, A::LinearMap)
-    m, n = size(A)
-    (n == size(X, 2) && m == size(Y, 2) && size(Y, 1) == size(X, 1)) ||
-        throw(DimensionMismatch("left mul!"))
+    mA, nA = size(A) # A always has two dimensions
+    mB, nB = size(B, 1), size(B, 2)
+    if mB != nA
+        throw(DimensionMismatch("left factor has dimensions ($mA,$nA), right factor has dimensions ($mB,$nB)"))
+    end
+    if size(C, 1) != mA || size(C, 2) != nB
+        throw(DimensionMismatch("result has dimensions $(size(C)), needs ($mA,$nB)"))
+    end
     return nothing
 end
 
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -4,7 +4,6 @@ export LinearMap
 export ⊗, kronsum, ⊕
 
 using LinearAlgebra
-import LinearAlgebra: AdjointAbsVec, TransposeAbsVec
 using SparseArrays
 
 if VERSION < v"1.2-"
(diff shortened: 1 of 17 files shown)


JeffFessler:
I found a bug in the tests.  Still working on it...

JeffFessler:
Ok, as I expected handling `*` was easy, but `mul!` is hard because it is difficult for me to follow the many variations of `mul!` throughout the package.  The main change is that most of the existing `mul!` methods had `AbstractVector` as the target, and this has to be generalized to a `Union` that includes Adjoint or Transpose abstract vectors.  (I called it `VecOut`.)  I ended up isolating most of the new code to `left.jl` in src/ and in test/.
Now I need some help.
Firstly, I can't get it to work for 1.0 and honestly I'd rather not spend time trying to make new features work for an old Julia version.  So my kludge is simply not to test the new features for old versions.
More importantly, the `mul!` works for several subtypes of LinearMaps, but not all subtypes, as seen in the notes in `test/left.jl`.  Some types end up in infinite loops leading to stack overflow.  (Multiple dispatch is powerful but also hard to trace what is happening...)  Probably some of the issues for those types is that they use `@eval` to create some of their methods and I was less sure how to modify those.

src/LinearMaps.jl
```diff
@@ -61,6 +66,13 @@ function check_dim_mul(Y::AbstractMatrix, A::LinearMap, X::AbstractMatrix)
     return nothing
 end
 
+# check dimension consistency for left multiplication x = y'*A
```
dkarrasch, line 69: What if we relax the "output" type of `mul!` to `LinearAlgebra.AbstractVecOrMat`, and make `check_dim_mul` check all the usual `size` conditions as in line 62 above? We would need only one
```julia
function check_dim_mul(Y::AbstractVecOrMat, A::LinearMap, X::AbstractVecOrMat)
    # @info "checked matrix dimensions" # uncomment for testing
    m, n = size(A)
    (m == size(Y, 1) && n == size(X, 1) && size(Y, 2) == size(X, 2)) || throw(DimensionMismatch("mul!"))
     return nothing
 end
```
First of all, I don't think these checks are expensive. And second, for `X,Y::AbstractVector`s, and `size(Y, 2) == size(X, 2))` may get constant-propagated and compiled away, since they return `1` in that case. Below, we could have `y::AbstractVecOrMat` in all the `mul!`s (which would make your `VecOut` obsolete), and we would need the `VecIn` only in a few cases.

This is not tested and probably not thought through well, but what came to my mind immediately. I would hope that we wouldn't have any issues continuing supporting Julia v1.0.

test/left.jl
```diff
@@ -0,0 +1,62 @@
+using Test, LinearMaps
+using LinearAlgebra: mul!
+
+# y'*L is an exception to the left multiplication rule that makes a WrappedMap
+
+@static if VERSION ≥ v"1.4.0"
+
+function left_tester(L::LinearMap{T}) where {T}
```
dkarrasch, line 8: ❤️ I love this "test function" style, and realized recently in the other PR when searching for a certain "test pattern" for quite a while that all our tests should be rewritten like that.

src/LinearMaps.jl
JeffFessler: Thanks for the reply and suggestions.  I will take a stab at that and get back to you when done or stuck.

JeffFessler:
After improving `check_dim_mul` as @dkarrasch suggested, I was able to track down the remaining bugs and greatly simplify things.  It turns out that the existing `AbstractVector` outputs were just fine for nearly all existing functions.  So now everything works for all `LinearMap` types and code review will be easier (only 6 files changed instead of 14).  And Julia 1.0 is 95% working now.  It works fine for `mul!(x, y', A)` but not for `mul!(x, transpose(y), A)`, oddly, so I have fenced out that test.  I kind of doubt that anyone will ever use `mul!(x, transpose(y), A)` in any Julia version, but at least I have isolated the issue to a small part in case you want to take a stab at it.  Obviously when the time comes I'd recommend squashing the commits because there were a lot of useless tries along the way...

dkarrasch:
The issue with v1.0 was minor. You forgot to define a 3-arg `mul!` for the `transpose(y)*A` case. I know this is poorly documented (and improving that is on my todo list), but I decided in one of the more recent PRs to clearly distinguish between 3-arg and 5-arg calls. By default, we should not direct 3-arg calls to 5-arg `mul!`s as in the matrix case, because that is, in general, a bad idea for `FunctionMap`s. I also know that this seems to blow up the amount of `mul!` methods, but I think the bigger issue about that is lack of documentation. The "repetitions" follow a clear pattern, and if that's well-communicated, it looks much less scary. Having said that, I wouldn't mind if we could reduce it again, but I was having big troubles with this "5-arg default" in #74.

From my point of view, this is ready for a final quick review, and then I'll merge and release.

dkarrasch:
> Thumbs up on the duck typing for `check_dim_mul` and the more informative error messages.

Yeah, I basically copied them from `LinearAlgebra.generic_matmul` or something like that.

dkarrasch:
Technically, this is a breaking change, right? So by the semver rules, this needs to get a new major version number. I'd suggest to put this together with #97 and perhaps #87 and make a v3.0 release?

I'm on vacation (offline) the next two weeks. Is there any hurry in getting this released?

JeffFessler:
Waiting is fine.  If needed I can use this branch in the meantime.

src/left.jl
```diff
@@ -11,32 +11,35 @@
...
-    @boundscheck check_dim_mul(x, y, A)
-    @inbounds mul!(adjoint(x), A', y', α, β)
-    return adjoint(x)
-end
+# multiplication with vector/matrix
+for Atype in (AdjointAbsVec, Adjoint{<:Any,<:AbstractMatrix})
+    @eval Base.@propagate_inbounds function LinearAlgebra.mul!(x::AbstractMatrix, y::$Atype, A::LinearMap)
+        @boundscheck check_dim_mul(x, y, A)
+        @inbounds mul!(adjoint(x), A', y')
+        return x
```
dkarrasch, line 22: I think returning `adjoint(x)` was actually wrong, so that `mul!` returned a "standing vector". The way Julia's `mul!` writes into the `adjoint` wrapper of `x` is so beautiful.

JeffFessler:
Lots of nice updates.
How close are we to a new tagged version with this left vector multiply merge in it?

PR #104, opened 2020-08-31
julia annotation

JeffFessler:
hopefully will help with syntax highlighting?


dkarrasch:
Thanks, Jeff. For small things like that, I guess it will be more convenient to use the "suggestion" feature in the code review web interface. Have you tried that? In case not: you click on the line number next to the diff, and then there is a +/- button or so. You can turn that on even manually, by writing three backticks followed by "suggestion". The following code box is then "offered" as a commit to the same PR. I can commit that to the PR, and it will keep your authorship. Let's try that over there, and I'll close the small PRs here. Okay?

JeffFessler:
sure, i didn't know about that option!  (authorship for one liners doesn't matter to me anyway but nice of you to mention it).

PR #105, opened 2020-08-31
add fill arrays

JeffFessler:
if you want - up to you.  i found it to be useful, though i admit i have not wrapped one in a LM yet...

### DIFF
--- a/docs/src/related.md
+++ b/docs/src/related.md
@@ -26,6 +26,8 @@ there exist further related packages in the Julia ecosystem:
 
 *   [`Kronecker.jl`](https://github.com/MichielStock/Kronecker.jl)
 
+*   [`FillArrays.jl`](https://github.com/JuliaArrays/FillArrays.jl)
+
 Since these packages provide types that are subtypes of Julia `Base`'s `AbstractArray` type,
 objects of those types can be wrapped by a `LinearMap` and freely mixed with, for instance,
 function-based linear maps.


dkarrasch:
I thought about this one, but I'm not sure we have functionality in LinearMaps.jl that does what FillArrays can do. So it feels like FillArrays.jl is not related (or the other way around? 😛) except that it provides an array type which we can wrap. But I assume there are plenty of such. Having said that, we could have a comment that emphasizes that one can wrap literally anything that subtypes to `AbstractArray`, in particular custom, non-`Base` matrix types.

JeffFessler:
they specifically provide a fast way of doing A*x when A is a FillArray so that's why i thought it was more relevant than generic abstractarray types, but it's no big deal either way.
actually, because `ones(m,n) = ones(m) * ones(n)'` we can easily do the same thing with a compositemap here so there's probably little to no value added to using a FillArray.

dkarrasch:
... except that we don't have a simple means to create a 1xm-"linear map" filled with a constant value. I like the idea of pointing out the generic wrapping mechanism, and then we could add a reference to the packages owned by the `JuliaArray` org or something, as an example.

JeffFessler:
oh right, it would take a functionmap with `sum`, and `fill` for the adjoint.

dkarrasch:
I didn't think about it that way, but that's indeed really cute. I wonder if one can achieve similar effects with this function-based definition of "FillMap"s... from what I understand, `FillArrays.jl` seems to allocate a filled array when calling `*(::Matrix,::Fill)`.

JeffFessler:
oops, seems that i may have misremembered about how fillarrays does A*x !
Closer inspection of
https://github.com/JuliaArrays/FillArrays.jl/blob/master/src/fillalgebra.jl
seems to have only some special cases, not the general case i was imagining.
well, i don't have time to help over there right now especially because the "axes" part of it is unfamiliar to me...

would the `FillMap` above be worth adding as an example in the notes?

dkarrasch:
I found your suggestion with `sum` and `fill` so cute that I immediately wrote up a little file with `FillMap` and methods. If that's relevant for someone, we could simply add it to the package. I haven't done any benchmarking yet, but I like the complete laziness of it. If not, then my little effort would not be wasted, because, as you say, we could include as an example in the "custom map" section.

dkarrasch:
By the way, it also implements the rectangular "`ZeroMap`" we were discussing at some point.

dkarrasch:
@JeffFessler Are you saying that this is worth being included in the package?

JeffFessler:
Oh, no strong view one way or another here.
I'd probably slight lean towards putting it as an example in the docs because I think it is it just a line or two of code, but I may be overlooking subtleties and couldn't find the "little file" you mentioned to see what it really looks like when fleshed out.

dkarrasch:
This is how it looks like when you want to integrate it with the other types etc.: https://github.com/Jutho/LinearMaps.jl/blob/dk/fillmap/src/fillmap.jl

JeffFessler:
neat!  my 1 line became 59 lines :)

line 14,15 revert -> `reverse` i think

now that i see this, maybe i'd recommend adding to the package and maybe adding a pointer in the documentation pointing to the code because this might be the simplest self-contained example of how to make a new LM subtype.
perhaps that is what you had in mind in "custom maps" (i wasn't sure).

dkarrasch:
Well, you could argue that your one line became an entire package with `FillArrays.jl`. 😛 What a code bloat! 🤣 

The last 15 lines are all about "reduction". You could safely omit those, as you would have to do if you simply defined it via a one-liner in a `FunctionMap`, and it would still work. Moreover, you could be lazy and let `MulStyle(::FillMap) = ThreeArg()`, the default. Then you don't even need to define the 5-arg `mul!`, and we're down to 25 lines. I'd argue that these reduction rules are what you get if you have a specific type, and that is what helps you get performance in the end.

I agree that this is a short self-contained example. And the `ThreeArg` vs. `FiveArg` issue can be nicely demonstrated with this type. So let's have both then, the type and a discussion in the documentation.

PR #108, opened 2020-09-10
Rework conversions, speed-up matrix Kronecker products

dkarrasch:
This should allow, for instance, to create a `Matrix` of desired `eltype`, or, as @JeffFessler would say, accuracy.

### DIFF
--- a/src/kronecker.jl
+++ b/src/kronecker.jl
@@ -119,10 +119,10 @@ end
     na, ma = size(At)
     mb, nb = size(B)
     Y = reshape(y, (mb, ma))
-    if na * mb * (ma + nb) < ma * nb * (na + mb) #nb*ma < mb*na
-        _unsafe_mul!(Y, B, convert(AbstractMatrix, X*At))
+    if nb*ma < mb*na 
+        _unsafe_mul!(Y, B, Matrix(X*At))
     else
-        _unsafe_mul!(Y, convert(AbstractMatrix, B*X), At isa MatrixMap ? At.lmap : At.λ)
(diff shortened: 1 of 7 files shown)


src/kronecker.jl
```diff
@@ -119,9 +119,9 @@ end
     na, ma = size(At)
     mb, nb = size(B)
     if nb*ma < mb*na
-        _unsafe_mul!(reshape(y, (mb, ma)), B, convert(Matrix, X*At))
+        _unsafe_mul!(reshape(y, (mb, ma)), B, convert(AbstractMatrix, X*At))
```
JeffFessler, line 122: i wish that somewhere near line 130 there would be a comment explaining what is unsafe here.
perhaps it is that the size of `y` is not checked?  if so, maybe a brief comment on why that's ok,
e.g., it is checked elsewhere?

lots of subtle stuff i don't quite get here so i can't offer much commentary on it.  seems ok...

dkarrasch: Yes, the `_unsafe` is due to lack of any size checking, as that is done (i) in `mul!` and (ii) during construction of the specific`LinearMap` types. I can add a comment in the main file to explain the global structure.

dkarrasch:
This improves matrix-based Kronecker products quite a bit. I believe that (though the approach is simple) we have the fastest code for (lazy-kron-matrix)-(vector) products. There are exceptions when matrices are so small that forming the Kronecker product explicitly is much faster (`Base`), and when you apply the KroneckerMap to wide matrices, which is not the primary scope of `LinearMaps.jl`.

PR #110, opened 2020-09-16
Fix _unsafe_mul! for custom maps

dkarrasch:
This fixes our new `_unsafe_mul!` mechanism for users with custom maps who define `mul!(y, A::TransposeMap{<:Any,<:CustomMap}, x)` (noticed while working on #111). Without the fix, one gets trapped by the generic `_unsafe_mul!(y, ::TransposeMap, x)`, and not redirected to the corresponding `mul!`. The effect is harmless, we may check sizes potentially an extra time.


dkarrasch:
I'll merge this. I don't think there is anything controversial here, just bugfixes.

PR #111, opened 2020-09-16
Add documentation on how to define custom maps

dkarrasch:


dkarrasch:
Comments welcome. One could turn the custom maps page into one generated by `Literate.jl` and let the stuff be actually computed during the generation step of the docs. That change would be almost no effort, and one could include "live output", if that's helpful by any means.

dkarrasch:
This may be of interest to @JeffFessler, @jagot, and @chriscoey. It would be great to have something helpful in the stable docs, once we release v3.0. Fine-tuning is always possible afterwards, of course.

docs/src/custom.md
```diff
@@ -1,3 +1,232 @@
...
+```julia
+function LinearAlgebra.mul!(y::AbstractVecOrMat, A::MyFillMap, x::AbstractVector)
+    LinearMaps.check_dim_mul(y, A, x)
+    return fill!(y, iszero(A.λ) ? zero(eltype(y)) : A.λ*sum(x))
+end
+```
+
+Again, due to generic fallbacks the following now "just work":
+
+* out-of-place multiplication `A*x`,
```
JeffFessler, line 47:
```suggest
* out-of-place multiplication by a vector on left or right: `A*x` and `y'*A`

Comment (dkarrasch) [reply docs/src/custom.md:47]: Well, within the flow of the help page it doesn't work yet, only after dealing with the transpose/adjoint further below. I'll add a comment there.

**Comment by dkarrasch** [comment]
This is so beautiful. I turned the `custom` page into a Literate page, and now the results of the code boxes are computed during the generation of the docs. I think this is ready to go.

**Review by JeffFessler — APPROVED** [review]
oops, oh right, of course.  belongs later.

**Comment by JeffFessler** [comment]
I looked at the docs and they look good.  One issue.  After the line:
"Not surprisingly, without further definitions, multipliying A' by x yields an error."
the (literate?) output is a vector, not an error as expected.

File: docs/src/custom.jl [hunk 2]
```diff
@@ -0,0 +1,201 @@
...
+# two *distinct* paths.
+
+# ### Path 1: Generic, non-invariant `LinearMap` subtypes
+
+# Generically, taking the transpose (or the adjoint) of a (real, resp.) map wraps the
+# linear map by a `TransposeMap`.
+
+typeof(A')
+
+# Not surprisingly, without further definitions, multipliying `A'` by `x` yields an error.
```
JeffFessler, line 141:
```suggest
# Not surprisingly, without further definitions, multiplying `A'` by `x` yields an error.

File: docs/src/custom.jl [hunk 3]
```diff
@@ -0,0 +1,201 @@
...
+
+A'x
```
JeffFessler, line 143: unsure why this is not erroring...

dkarrasch: Thanks for pointing this out. It doesn't error because I have defined `issymmetric` for `MyFillMap`, and because it realizes that my specific `A` is symmetric, it doesn't require the formal transpose, but uses `A` itself! I must have rearranged things and used to have the `issymmetric` definitions below this line. That's instructive, I think it's going to error if I make it nonsquare, or complex.

dkarrasch: Somehow Documenter doesn't show the error. Let's see if one can do something about it: https://github.com/JuliaDocs/Documenter.jl/issues/1420
Otherwise, I'll just type the error message manually.

PR #113, opened 2020-10-05
Add FillMaps

dkarrasch:
- [x] adjust the documentation & write a docstring

- [x] write a `show` method specialization (after #87 is included)

- [x] decide on constructor

### DIFF
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -247,11 +247,14 @@ include("show.jl") # show methods for LinearMap objects
     LinearMap(A::LinearMap; kwargs...)::WrappedMap
     LinearMap(A::AbstractMatrix; kwargs...)::WrappedMap
     LinearMap(J::UniformScaling, M::Int)::UniformScalingMap
+    LinearMap(λ::Number, M::Int, N::Int) = FillMap(λ, (M, N))::FillMap
+    LinearMap(λ::Number, dims::Dims{2}) = FillMap(λ, dims)::FillMap
     LinearMap{T=Float64}(f, [fc,], M::Int, N::Int = M; kwargs...)::FunctionMap
 
 Construct a linear map object, either from an existing `LinearMap` or `AbstractMatrix` `A`,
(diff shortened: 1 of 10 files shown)


src/LinearMaps.jl
```diff
@@ -270,6 +271,8 @@ LinearMap(f, M::Int; kwargs...) = LinearMap{Float64}(f, M; kwargs...)
 LinearMap(f, M::Int, N::Int; kwargs...) = LinearMap{Float64}(f, M, N; kwargs...)
 LinearMap(f, fc, M::Int; kwargs...) = LinearMap{Float64}(f, fc, M; kwargs...)
 LinearMap(f, fc, M::Int, N::Int; kwargs...) = LinearMap{Float64}(f, fc, M, N; kwargs...)
+LinearMap(λ::Number, M::Int, N::Int) = FillMap(λ, (M, N))
+LinearMap(λ::Number, dims::Dims{2}) = FillMap(λ, dims)
```
dkarrasch, line 275: I think the only somewhat controversial things are these constructors. They are formally unambiguous, but are they unambiguous enough for users (sufficiently distinct from `UniformScalinMap` constructor?)? Or should we export the `FillMap` construct and not include these?

Jutho: As this is one of few the `LinearMap` types that you want to construct explicitly, I guess it would make sense to export the `FillMap` constructor explicitly. It's somewhat asymmetric that `FunctionMap` is then still constructed via `LinearMap`. Just using `LinearMap(scalar, ...)` is also fine by me, no strong preference.

I do wonder what this is used for though :-) ?

dkarrasch: I don't know either what this is used for, but (a) there is an entire package (`FillArrays.jl`) devoted to such arrays, and (b) I do notice that our simple imlementation of matvecmul and matmatmul is even faster than theirs.

And yes, I agree that this is so specific that it is perhaps better to have an explicit constructor. It was more "sentimentality" that made me consider keeping the "one constructor for all types" status.

PR #114, opened 2020-10-07
Final piece for v3.0

### LINKED ISSUE #101: Towards v3.0
We have currently two PRs in the pipeline that are breaking and therefore require a major release. So I thought we should collect ideas on what to put in there to not have major release every month or so. Clearly, we will have

- [x] #100 

- [x] #97 

Furthermore, I'd suggest to include
- [x] #87 with the new show methods (I guess the only way to really get a feeling for whether it's good or not is to start using it and collect "annoyed" feedback, otherwise I feel this is too distant to form an opinion just by reviewing the PR)
- [x] a nice Documenter documentation, which then could nicely address #98 (That shouldn't be too hard: I'm planning to mainly cut our long README into more digestible pieces as a first approach.)

Anything else? Of course, if some hot PR comes in, there's nothing that stops us from merging it and releasing another v2.x. I'd hope to finish this some time in late August/early September. Any contributions/PRs/suggestions are more than welcome.

Have a good (rest of the) summer everyone!

EDIT:
- [x] I think we should relax the type of the output variable in the `mul!`s to `AbstractVecOrMat`. The following works perfectly fine with matrices in `LinearAlgebra`:
```
using LinearAlgebra
mul!(rand(5,1), rand(5,5), rand(5))
```

TODO:
- [x] overhaul by @Jutho 
- [x] document how to write custom map types #111 
- [ ] ~~document conversion logic~~ (not sure that's of general interest)
- [x] `FillMap`s

### LINKED ISSUE #109: Stable docs doesn't exist
Unfortunately there are no stable docs for your package. 
It seems like you added the ssh key for Documenter only recently. If you want to have stable docs a new version needs to be registered or you can build the docs locally and push to the `gh-pages` branch.

### LINKED ISSUE #120: Example code throws MethodError
Hi,
I'm running this on Julia-1.5:
```julia
using LinearMaps, LinearAlgebra

struct MyFillMap{T} <: LinearMaps.LinearMap{T}
    λ::T
    size::Dims{2}
    function MyFillMap(λ::T, dims::Dims{2}) where {T}
        all(≥(0), dims) || throw(ArgumentError("dims of MyFillMap must be non-negative"))
        promote_type(T, typeof(λ)) == T || throw(InexactError())
        return new{T}(λ, dims)
    end
end

Base.size(A::MyFillMap) = A.size

function LinearAlgebra.mul!(y::AbstractVecOrMat, A::MyFillMap, x::AbstractVector)
    LinearMaps.check_dim_mul(y, A, x)
    return fill!(y, iszero(A.λ) ? zero(eltype(y)) : A.λ*sum(x))
end

A = MyFillMap(5.0, (3, 3)); x = ones(3); sum(x)
```

But then multiplying throws an error:
```julia
A*x

MethodError: no method matching A_mul_B!(::Array{Float64,1}, ::Main.workspace11.MyFillMap{Float64}, ::Array{Float64,1})

Closest candidates are:

A_mul_B!(::AbstractArray{T,1} where T, !Matched::AbstractArray{T,2} where T, ::AbstractArray{T,1} where T) at /Users/ralph/.julia/packages/LinearMaps/07z1W/src/LinearMaps.jl:117

A_mul_B!(::AbstractArray{T,1} where T, !Matched::LinearMaps.WrappedMap, ::AbstractArray{T,1} where T) at /Users/ralph/.julia/packages/LinearMaps/07z1W/src/wrappedmap.jl:40

A_mul_B!(::AbstractArray{T,1} where T, !Matched::LinearMaps.UniformScalingMap, ::AbstractArray{T,1} where T) at /Users/ralph/.julia/packages/LinearMaps/07z1W/src/uniformscalingmap.jl:91

...

*(::Main.workspace11.MyFillMap{Float64}, ::Array{Float64,1})@LinearMaps.jl:74
top-level scope@Local: 1[inlined]
```

dkarrasch:
Let's have this for all the final details and for releasing v3.0.

Closes #109, closes #120, closes #101.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LinearMaps"
 uuid = "7a12625a-238d-50fd-b39a-03d52299707e"
-version = "2.7.0"
+version = "3.0"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/docs/src/custom.jl
+++ b/docs/src/custom.jl
@@ -200,7 +200,7 @@ mul!(similar(x)', x', A)
 # ## Application to matrices
 
 # By default, applying a `LinearMap` `A` to a matrix `X` via `A*X` does
-# *not* aplly `A` to each column of `X` viewed as a vector, but interprets
+# *not* apply `A` to each column of `X` viewed as a vector, but interprets
 # `X` as a linear map, wraps it as such and returns `(A*X)::CompositeMap`.
 # Calling the in-place multiplication function `mul!(Y, A, X)` for matrices,
 # however, does compute the columnwise action of `A` on `X` and stores the
--- a/docs/src/history.md
+++ b/docs/src/history.md
@@ -1,56 +1,84 @@
-# What's new?
+# Version history
 
-### What's new in v2.7
-*   Potential reduction of memory allocations in multiplication of
-    `LinearCombination`s, `BlockMap`s, and real- or complex-scaled `LinearMap`s.
-    For the latter, a new internal type `ScaledMap` has been introduced.
-*   Multiplication code for `CompositeMap`s has been refactored to facilitate to
-    provide memory for storage of intermediate results by directly calling helper
-    functions.
+## What's new in v3.0
 
-### What's new in v2.6
-*   New feature: "lazy" Kronecker product, Kronecker sums, and powers thereof
-    for `LinearMap`s. `AbstractMatrix` objects are promoted to `LinearMap`s if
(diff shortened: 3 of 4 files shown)


Jutho:
I will try to review 3.0 by the end of the week.

Jutho:
By the way, over at https://quantumghent.github.io/software/ , one of my postdocs went through the trouble of making nice Julia-inspired logos for all the packages we have been working on in our group. LinearMaps.jl is still missing a logo though, also due to lack of inspiration. The one for KrylovKit is also still under debate.

Hence, any inspiration is welcome.

PR #119, opened 2020-12-05
WIP: review before version 3

Jutho:
Part 1, more to go (tomorrow or the days after that).

### DIFF
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -186,9 +186,6 @@ function Base.hvcat(rows::Tuple{Vararg{Int}},
                 na = size(As[j+k], 1)
                 ni >= 0 && ni != na &&
                     throw(DimensionMismatch("mismatch in number of rows"))
-                # here we check all maps, in hcat and vcat we only find the first map
-                # which is not UniformScaling to set the size.
-                # Should we also include this check in hcat and vcat?
                 ni = na
             end
         end
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -1,19 +1,26 @@
-struct BlockMap{T,As<:Tuple{Vararg{LinearMap}},Rs<:Tuple{Vararg{Int}},Rranges<:Tuple{Vararg{UnitRange{Int}}},Cranges<:Tuple{Vararg{UnitRange{Int}}}} <: LinearMap{T}
+struct BlockMap{T,
+                As<:LinearMapTuple,
+                Rs<:Tuple{Vararg{Int}},
+                Rranges<:Tuple{Vararg{UnitRange{Int}}},
+                Cranges<:Tuple{Vararg{UnitRange{Int}}}} <: LinearMap{T}
     maps::As
     rows::Rs
     rowranges::Rranges
     colranges::Cranges
-    function BlockMap{T,R,S}(maps::R, rows::S) where {T, R<:Tuple{Vararg{LinearMap}}, S<:Tuple{Vararg{Int}}}
-        for n in eachindex(maps)
-            A = maps[n]
-            @assert promote_type(T, eltype(A)) == T "eltype $(eltype(A)) cannot be promoted to $T in BlockMap constructor"
+    function BlockMap{T,As,Rs}(maps::As, rows::Rs) where
+                {T, As<:LinearMapTuple, Rs<:Tuple{Vararg{Int}}}
+        for TA in Base.Generator(eltype, maps)
+            promote_type(T, TA) == T ||
+                error("eltype $TA cannot be promoted to $T in BlockMap constructor")
         end
         rowranges, colranges = rowcolranges(maps, rows)
-        return new{T,R,S,typeof(rowranges),typeof(colranges)}(maps, rows, rowranges, colranges)
+        Rranges, Cranges = typeof(rowranges), typeof(colranges)
+        return new{T, As, Rs, Rranges, Cranges}(maps, rows, rowranges, colranges)
     end
 end
 
-BlockMap{T}(maps::As, rows::S) where {T,As<:Tuple{Vararg{LinearMap}},S} = BlockMap{T,As,S}(maps, rows)
+BlockMap{T}(maps::As, rows::Rs) where {T, As<:LinearMapTuple, Rs} =
+    BlockMap{T, As, Rs}(maps, rows)
 
 MulStyle(A::BlockMap) = MulStyle(A.maps...)
 
@@ -27,24 +34,26 @@ map in `maps`, according to its position in a virtual matrix representation of t
 block linear map obtained from `hvcat(rows, maps...)`.
 """
 function rowcolranges(maps, rows)
-    rowranges = ()
-    colranges = ()
+    rowranges = ntuple(n->1:0, Val(length(rows)))
+    colranges = ntuple(n->1:0, Val(length(maps)))
     mapind = 0
     rowstart = 1
-    for row in rows
-        xinds = vcat(1, map(a -> size(a, 2), maps[mapind+1:mapind+row])...)
-        cumsum!(xinds, xinds)
+    for (i, row) in enumerate(rows)
         mapind += 1
         rowend = rowstart + size(maps[mapind], 1) - 1
-        rowranges = (rowranges..., rowstart:rowend)
-        colranges = (colranges..., xinds[1]:xinds[2]-1)
+        rowranges = Base.setindex(rowranges, rowstart:rowend, i)
+        colstart = 1
+        colend = size(maps[mapind], 2)
+        colranges = Base.setindex(colranges, colstart:colend, mapind)
         for colind in 2:row
-            mapind +=1
-            colranges = (colranges..., xinds[colind]:xinds[colind+1]-1)
+            mapind += 1
+            colstart = colend + 1
+            colend += size(maps[mapind], 2)
+            colranges = Base.setindex(colranges, colstart:colend, mapind)
         end
         rowstart = rowend + 1
     end
-    return rowranges::NTuple{length(rows), UnitRange{Int}}, colranges::NTuple{length(maps), UnitRange{Int}}
+    return rowranges, colranges
 end
 
 Base.size(A::BlockMap) = (last(last(A.rowranges)), last(last(A.colranges)))
@@ -71,7 +80,7 @@ julia> L * ones(Int, 6)
  6
 ```
 """
-function Base.hcat(As::Union{LinearMap,UniformScaling,AbstractVecOrMat}...)
+function Base.hcat(As::Union{LinearMap, UniformScaling, AbstractVecOrMat}...)
     T = promote_type(map(eltype, As)...)
     nbc = length(As)
 
@@ -83,7 +92,9 @@ function Base.hcat(As::Union{LinearMap,UniformScaling,AbstractVecOrMat}...)
             break
         end
     end
-    nrows == -1 && throw(ArgumentError("hcat of only UniformScaling objects cannot determine the linear map size"))
+    errmsg = "hcat of only UniformScaling objects cannot determine the linear map size"
+    nrows == -1 && throw(ArgumentError(errmsg))
+    # does this mean we still have type piracy? I vagely remember this being discussed
     return BlockMap{T}(promote_to_lmaps(fill(nrows, nbc), 1, 1, As...), (nbc,))
 end
 
@@ -124,8 +135,9 @@ function Base.vcat(As::Union{LinearMap,UniformScaling,AbstractVecOrMat}...)
             break
         end
     end
-    ncols == -1 && throw(ArgumentError("vcat of only UniformScaling objects cannot determine the linear map size"))
-
+    errmsg = "vcat of only UniformScaling objects cannot determine the linear map size"
+    ncols == -1 && throw(ArgumentError(errmsg))
+    # same question about type piracy
     return BlockMap{T}(promote_to_lmaps(fill(ncols, nbr), 1, 2, As...), ntuple(i->1, nbr))
 end
 
@@ -160,10 +172,12 @@ julia> L * ones(Int, 6)
 """
 Base.hvcat
 
-function Base.hvcat(rows::Tuple{Vararg{Int}}, As::Union{LinearMap,UniformScaling,AbstractVecOrMat}...)
+function Base.hvcat(rows::Tuple{Vararg{Int}},
+                    As::Union{LinearMap, UniformScaling, AbstractVecOrMat}...)
     nr = length(rows)
     T = promote_type(map(eltype, As)...)
-    sum(rows) == length(As) || throw(ArgumentError("mismatch between row sizes and number of arguments"))
+    sum(rows) == length(As) ||
+        throw(ArgumentError("mismatch between row sizes and number of arguments"))
     n = fill(-1, length(As))
     j = 0
     for i in 1:nr # infer UniformScaling sizes from row counts, if possible:
@@ -173,13 +187,14 @@ function Base.hvcat(rows::Tuple{Vararg{Int}}, As::Union{LinearMap,UniformScaling
                 na = size(As[j+k], 1)
                 ni >= 0 && ni != na &&
                     throw(DimensionMismatch("mismatch in number of rows"))
+                # here we check all maps, in hcat and vcat we only find the first map
+                # which is not UniformScaling to set the size.
+                # Should we also include this check in hcat and vcat?
                 ni = na
             end
         end
         if ni >= 0
-            for k = 1:rows[i]
-                n[j+k] = ni
-            end
+            n[j .+ (1:rows[i])] .= ni
         end
         j += rows[i]
     end
@@ -218,7 +233,8 @@ function check_dim(A, dim, n)
 end
 
 promote_to_lmaps_(n::Int, dim, A::AbstractMatrix) = (check_dim(A, dim, n); LinearMap(A))
-promote_to_lmaps_(n::Int, dim, A::AbstractVector) = (check_dim(A, dim, n); LinearMap(reshape(A, length(A), 1)))
+promote_to_lmaps_(n::Int, dim, A::AbstractVector) =
+    (check_dim(A, dim, n); LinearMap(reshape(A, length(A), 1)))
 promote_to_lmaps_(n::Int, dim, J::UniformScaling) = UniformScalingMap(J.λ, n)
 promote_to_lmaps_(n::Int, dim, A::LinearMap) = (check_dim(A, dim, n); A)
 promote_to_lmaps(n, k, dim) = ()
@@ -236,16 +252,18 @@ function isblocksquare(A::BlockMap)
     return all(==(N), rows)
 end
 
+symindex(i, N) = ((k, l) = divrem(i-1, N); return k + l * N + 1)
+
 # the following rules are sufficient but not necessary
 function LinearAlgebra.issymmetric(A::BlockMap)
     isblocksquare(A) || return false
     N = length(A.rows)
     maps = A.maps
-    symindex = vec(permutedims(reshape(collect(1:N*N), N, N)))
     for i in 1:N*N
-        if (i == symindex[i] && !issymmetric(maps[i]))
+        isym = symindex(i, N)
+        if (i == isym && !issymmetric(maps[i]))
             return false
-        elseif (maps[i] != transpose(maps[symindex[i]]))
+        elseif (maps[i] != transpose(maps[isym]))
             return false
         end
     end
@@ -257,11 +275,11 @@ function LinearAlgebra.ishermitian(A::BlockMap)
     isblocksquare(A) || return false
     N = length(A.rows)
     maps = A.maps
-    symindex = vec(permutedims(reshape(collect(1:N*N), N, N)))
     for i in 1:N*N
-        if (i == symindex[i] && !ishermitian(maps[i]))
+        isym = symindex(i, N)
+        if (i == isym && !ishermitian(maps[i]))
             return false
-        elseif (maps[i] != adjoint(maps[symindex[i]]))
+        elseif (maps[i] != adjoint(maps[isym]))
             return false
         end
     end
@@ -272,7 +290,8 @@ end
 # comparison of BlockMap objects, sufficient but not necessary
 ############
 
-Base.:(==)(A::BlockMap, B::BlockMap) = (eltype(A) == eltype(B) && A.maps == B.maps && A.rows == B.rows)
+Base.:(==)(A::BlockMap, B::BlockMap) =
+    (eltype(A) == eltype(B) && A.maps == B.maps && A.rows == B.rows)
 
 ############
 # multiplication helper functions
@@ -299,7 +318,7 @@ function ___blockmul!(y, A, x, α, β, ::Nothing)
         mapind += 1
         _unsafe_mul!(yrow, maps[mapind], selectdim(x, 1, xinds[mapind]), α, β)
         for _ in 2:row
-            mapind +=1
+            mapind += 1
             _unsafe_mul!(yrow, maps[mapind], selectdim(x, 1, xinds[mapind]), α, true)
         end
     end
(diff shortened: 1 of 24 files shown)


src/uniformscalingmap.jl
```diff
@@ -58,41 +60,27 @@ end
...
+    elseif isone(λ)
+        iszero(β) && return y .= x .* α
+        isone(β) && return y .+= x .* α
+        return y .= y .* β .+ x .* α
+    else
+        iszero(β) && return y .= λ .* x .* α
+        isone(β) && return y .+= λ .* x .* α
+        return y .= y .* β .+ λ .* x .* α
+    end
+end
```
dkarrasch, line 83: This was originally so "complicated", because we wanted to save the multiplication by `λ` in case it's 1. But you don't seem to think that's relevant? I'm fine with that, just wanted to make sure it has been said and considered.

src/transpose.jl
```diff
@@ -52,54 +52,52 @@ Base.:(==)(A::LinearMap, B::AdjointMap)      = ishermitian(A) && B.lmap == A
...
+_unsafe_mul!(y::AbstractVecOrMat, A::AdjointMap, x::AbstractVector, α::Number, β::Number) =
+    ishermitian(A.lmap) ?
+        _unsafe_mul!(y, A.lmap, x, α, β) : _generic_mapvec_mul!(y, A, x, α, β)
+_unsafe_mul!(y::AbstractMatrix, A::AdjointMap, x::AbstractMatrix, α::Number, β::Number) =
+    ishermitian(A.lmap) ?
+        _unsafe_mul!(y, A.lmap, x, α, β) : _generic_mapmat_mul!(y, A, x, α, β)
+
 # # ConjugateMap
-for (intype, outtype) in ((AbstractVector, AbstractVecOrMat), (AbstractMatrix, AbstractMatrix))
+const ConjugateMap = AdjointMap{<:Any, <:TransposeMap}
```
dkarrasch, line 82: Beautiful!!! I had been thinking about introducing such a map type, but apparently one can get away with a simple `const`. Wonderful.

src/transpose.jl
```diff
@@ -52,54 +52,52 @@ Base.:(==)(A::LinearMap, B::AdjointMap)      = ishermitian(A) && B.lmap == A
...
+# canonical order of adjoint followed by transpose
+LinearAlgebra.transpose(A::AdjointMap) = adjoint(transpose(A.lmap))
```
dkarrasch, line 84: Maybe add
```suggestion
LinearAlgebra.transpose(A::AdjointMap) = adjoint(transpose(A.lmap))
LinearAlgebra.transpose(A::ConjugateMap) = adjoint(A.lmap.lmap)
```
to keep the types short?

src/left.jl
```diff
@@ -50,6 +50,7 @@ end
 function mul!(x::AbstractMatrix, y::AdjointAbsVecOrMat, A::LinearMap, α::Number, β::Number)
     check_dim_mul(x, y, A)
     _unsafe_mul!(x', conj(α)*A', y', true, conj(β))
+    # why not?: _unsafe_mul!(x', A', y', conj(α), conj(β))
```
dkarrasch, line 53: This is because 5-arg `mul!` always interprets (and performs) the multiplication by `α` from the right. But when we transpose/adjoint the whole product, the order should be formally reversed, and for non-commutative types that order matters. But this is of no harm, actually, for `RealOrComplex`: in this case, `conj(α)*A'` creates a `ScaledMap`

https://github.com/Jutho/LinearMaps.jl/blob/b38c1ac1900d4751237bbef62927477017f4f7d0/src/scaledmap.jl#L30

and `mul!` for `ScaledMap`s does exactly what you propose:

https://github.com/Jutho/LinearMaps.jl/blob/b38c1ac1900d4751237bbef62927477017f4f7d0/src/scaledmap.jl#L46-L55

dkarrasch (commented):
Great to see this started!

src/uniformscalingmap.jl
Jutho: The `isone(λ)` case is handled above no?

src/transpose.jl
Jutho: Yes good idea

src/uniformscalingmap.jl
dkarrasch: You're right. All the lines basically look so similar, so I overlooked that. 😅

src/composition.jl
```diff
@@ -117,30 +118,40 @@ function Base.:(*)(A₁::CompositeMap, A₂::CompositeMap)
...
+# end
+# function _unsafe_mul!(y::AbstractVecOrMat, A::CompositeMap, x::AbstractVector)
+#     _compositemul!(y, A, x, similar(y, size(A.maps[1], 1)), similar(y, size(A.maps[2], 1)))
+# end
+_unsafe_mul!(y::AbstractVecOrMat, A::CompositeMap, x::AbstractVector) =
+    _compositemul!(y, A, x)
 
-function _compositemul!(y::AbstractVecOrMat, A::CompositeMap{<:Any,<:Tuple{LinearMap,LinearMap}}, x::AbstractVector, z::AbstractVector)
+function _compositemul!(y::AbstractVecOrMat, A::CompositeMap2, x::AbstractVector)
+    # was there any advantage in having this z an argument, this is not a public method?
```
dkarrasch, line 144: Yes, this was a service to external packages to be able to hook into 2-map CompositeMaps and provide a preallocated intermediate vector (@JeffFessler asked for it). I realize we should have tested that specifically, to make clear we want that. Similarly for higher-order products.

src/functionmap.jl
```diff
@@ -17,20 +17,26 @@ function FunctionMap{T}(f::F1, fc::F2, M::Int, N::Int;
...
+    FunctionMap{T}(f, nothing, M, M; kwargs...)
+FunctionMap{T}(f, M::Int, N::Int; kwargs...) where {T} =
+    FunctionMap{T}(f, nothing, M, N; kwargs...)
+FunctionMap{T}(f, fc, M::Int; kwargs...) where {T} =
+    FunctionMap{T}(f, fc, M, M; kwargs...)
 
 # properties
 Base.size(A::FunctionMap) = (A.M, A.N)
-Base.parent(A::FunctionMap) = (A.f, A.fc)
+Base.parent(A::FunctionMap) = (A.f, A.fc) # is this meanigful/useful?
```
dkarrasch, line 29: IIRC the reason for `parent` was for manual inspection of wrapped maps, because `show` is now rather sparse. It may be used in the package only rarely, though.

src/scaledmap.jl
```diff
@@ -15,8 +15,8 @@ ScaledMap(λ::S, lmap::A) where {S<:RealOrComplex,A<:LinearMap} =
 Base.size(A::ScaledMap) = size(A.lmap)
 Base.parent(A::ScaledMap) = (A.λ, A.lmap)
 Base.isreal(A::ScaledMap) = isreal(A.λ) && isreal(A.lmap)
-LinearAlgebra.issymmetric(A::ScaledMap) = issymmetric(A.lmap)
-LinearAlgebra.ishermitian(A::ScaledMap) = ishermitian(A.lmap)
+LinearAlgebra.issymmetric(A::ScaledMap) = isreal(A.λ) && issymmetric(A.lmap)
+LinearAlgebra.ishermitian(A::ScaledMap) = isreal(A.λ) && ishermitian(A.lmap)
```
dkarrasch, line 19: Why is that? `A.λ` is just a constant factor. If it's complex,  it shouldn't contradict symmetry?

src/composition.jl
```diff
@@ -1,18 +1,18 @@
...
         N = length(maps)
         for n in 2:N
             check_dim_mul(maps[n], maps[n-1])
         end
-        for n in eachindex(maps)
-            A = maps[n]
-            @assert promote_type(T, eltype(A)) == T "eltype $(eltype(A)) cannot be promoted to $T in CompositeMap constructor"
+        for TA in Base.Generator(eltype, maps)
+            promote_type(T, TA) == T ||
+                error("eltype $TA cannot be promoted to $T in CompositeMap constructor")
```
dkarrasch, line 10: That's clever! That makes the loop body depend on types, and avoids closure about `maps`, right?

src/functionmap.jl
JeffFessler: perhaps any parent-like method would be most useful if it returned all the arguments needed to recreate the object.  at a minimum that would include size too.  maybe a named tuple would be more useful too.
not sure i would use it much though.

src/composition.jl
JeffFessler: i'm still hoping to exploit this at some point.  kind of waiting for v3.0...

src/scaledmap.jl
JeffFessler: sqrt(-1) * I is not Hermitian.  the diagonal elements must be real.  i agree with jutho here.

src/composition.jl
JeffFessler: if @dkarrasch has to ask about this, then others like me will have even less chance of understanding so this would be a nice place to add a comment in the code...

src/scaledmap.jl
Jutho: Right, it's probably not needed for `issymmetric` if you are happy with complex symmetric (which we are). For `ishermitian`, it is necessary though.

src/functionmap.jl
Jutho: From Base:
```
help?> Base.parent
  parent(A)

  Return the underlying "parent array”. ...
```
So I see this making sense for `WrappedMap`, but other than that I am not entirely sure. Where is this used?

src/composition.jl
Jutho: Sure I am happy to revert this change. One suggestion could be to make temporary memory buffers default arguments, with the proper initialization as default value. What do you think of that suggestion @dkarrasch and @JeffFessler , I am fine either way.

Jutho: Mostly, the motivation of my change was to have shorter lines and avoid tricky linebreaking situations

src/functionmap.jl
dkarrasch: I think we use `parent` only somewhere in `kronecker.jl`, and then potentially in an inconsistent way. It's not obvious what `parent` should return given the wide range of `LinearMap` subtypes. I'd be fine with actually removing it. I introduced this when I cut down the `show` method to show much less details, and then I used it in one or two places as it was handy. But this is not really crucial.

src/composition.jl
Jutho: In Julia 1.x there is now `Base.Iterators.map`, a lazy map, which does exactly this, it just calls `Base.Generator`. `Base.Generator` is what is being called by the generator syntax. It never creates a closure over the iterator, only over other arguments provided in the function. Say you have somewhere
```
f(x, a) for x in b
```
that basically lowers to `Base.Generator(x->f(x,a), b)` and thus creates a closure over/of/about (what is the correct preposition?) `a`.

Jutho: I will add a comment.

src/composition.jl
dkarrasch: You mean like positional arguments with a default value, that can be overwritten? That sounds like a good idea. We would just make sure that we don't allocate two vectors for a `CompositeMap1`.

Jutho:
blockmap was a though one, as i was not familiar with the code. Kronecker will have to wait for another day (hopefully tomorrow).

src/blockmap.jl
```diff
@@ -173,13 +187,14 @@ function Base.hvcat(rows::Tuple{Vararg{Int}}, As::Union{LinearMap,UniformScaling
                 na = size(As[j+k], 1)
                 ni >= 0 && ni != na &&
                     throw(DimensionMismatch("mismatch in number of rows"))
+                # here we check all maps, in hcat and vcat we only find the first map
+                # which is not UniformScaling to set the size.
+                # Should we also include this check in hcat and vcat?
```
dkarrasch, line 192: This check is done in `promote_to_lmaps_` in the new line 141 above.

dkarrasch (commented):
Thanks @Jutho!

src/blockmap.jl
```diff
@@ -83,7 +92,9 @@ function Base.hcat(As::Union{LinearMap,UniformScaling,AbstractVecOrMat}...)
             break
         end
     end
-    nrows == -1 && throw(ArgumentError("hcat of only UniformScaling objects cannot determine the linear map size"))
+    errmsg = "hcat of only UniformScaling objects cannot determine the linear map size"
+    nrows == -1 && throw(ArgumentError(errmsg))
+    # does this mean we still have type piracy? I vagely remember this being discussed
```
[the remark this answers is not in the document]
Jutho: Ok great. Annoying that this does not work for `cat`. Looking into the implementation of `cat` in Base, LinearAlgebra and SparseArrays, it seems like a complete mess. Something like
```
cat(rand(Float32, (2,2)), rand(Float64, (3,3)); dims = (1,2))
```
goes via SparseArrays. Anyway, looking at SparseArrays, it seems like we might want to overload `Base._cat` instead of `Base.cat`?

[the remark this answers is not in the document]
Jutho: Also, for the above, it means that it is impossible to trigger the error message?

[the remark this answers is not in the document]
dkarrasch: > it seems like we might want to overload Base._cat instead of Base.cat

If that does the job, that seems fine to me.

> Also, for the above, it means that it is impossible to trigger the error message?

Yes, indeed. Good point!

src/composition.jl
JeffFessler: I very much like the idea of temp buffers as default arguments (positional or keyword, either ok) because my hunch is that will simplify what i hope to do with these.

src/blockmap.jl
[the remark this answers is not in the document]
JeffFessler: too bad that `?Base._cat` returns nothing useful.  sigh.

src/kronecker.jl
```diff
@@ -46,24 +44,30 @@ julia> Matrix(Δ)
...
+    KroneckerMap{promote_type(eltype(A), eltype(B))}((A, B))
+Base.kron(A::LinearMap, B::KroneckerMap) =
+    KroneckerMap{promote_type(eltype(A), eltype(B))}(tuple(A, B.maps...))
+Base.kron(A::KroneckerMap, B::LinearMap) =
+    KroneckerMap{promote_type(eltype(A), eltype(B))}(tuple(A.maps..., B))
+Base.kron(A::KroneckerMap, B::KroneckerMap) =
+    KroneckerMap{promote_type(eltype(A), eltype(B))}(tuple(A.maps..., B.maps...))
+Base.kron(A::LinearMap, B::LinearMap, Cs::LinearMap...) =
+    KroneckerMap{promote_type(eltype(A), eltype(B), map(eltype, Cs)...)}(tuple(A, B, Cs...))
+    # could this just be a recursive definition: kron(A, B, C) = kron(kron(A, B), C) ?
```
dkarrasch, line 57: One could try. I wanted to make sure in general that when you have first a couple of matrices and only eventually a `LinearMap`, then we don't want to `kron` the matrices first, but we want it all lazy. But I think we have a test for that. But my guess is that I had tried that already (in fact, it's the generic base behaviour).

src/kronecker.jl
```diff
@@ -181,20 +193,20 @@ end
 ###############
 # KroneckerSumMap
 ###############
-struct KroneckerSumMap{T, As<:Tuple{LinearMap,LinearMap}} <: LinearMap{T}
+struct KroneckerSumMap{T, As<:Tuple{LinearMap, LinearMap}} <: LinearMap{T}
     maps::As
-    function KroneckerSumMap{T, As}(maps::As) where {T, As}
-        for n in eachindex(maps)
-            A = maps[n]
-            size(A, 1) == size(A, 2) || throw(ArgumentError("operators need to be square in Kronecker sums"))
```
dkarrasch, line 189: 🤦

dkarrasch (commented):
👍

src/kronecker.jl
Jutho: Oh right, I did not think about that. Although wait, this is a method for `LinearMap` objects only, right, so the recursive definition will just end up calling `kron(::LinearMap, ::LinearMap)` which create a `KroneckerMap`, no?

src/kronecker.jl
Jutho: Is there a reason actually that `KroneckerSum` is restricted to two arguments. If it is like the logarithm of a Kronecker product, it could also have an arbitrary number of factors, or rather, terms, right? I didn't know this was a thing even until you implemented it, although I do encounter it also in my research field.

JeffFessler: This question made me curious, so I have just verified that it is associative.  (It is not commutative.)
So it can extend to more arguments, though I think the easiest construction is probably recursive.
Not to much about this that I could find online other than this:
https://www.euclideanspace.com/maths/algebra/multidimensional/kroneckerSum/index.htm 
And that is an unreliable source because it wrongly claims commutativity...

Jutho:
I've left some more questions. I am otherwise finished; this package has evolved very nicely and has become very mature. I like the show implementation, might mimic this design in other packages where a similar structure could be useful.

I did not review the tests, but given the high coverage, these must be great!

src/kronecker.jl
dkarrasch: True. I didn't remember because it wasn't shown in the diff, but making sure that, even when you have several matrices first, that those are lazily Kronecker multiplied, is the task of the "symbolic loop" below. So, in fact, this `A, B, Cs...` method could be removed, it seems.

src/kronecker.jl
dkarrasch: Surely, I must have been considering this, given the KroneckerMap design above. You can `kronsum` several maps, but for implementation reasons that I don't remember, for instance, `kronsum(A, B, C) = kronsum(kronsum(A, B), C)` simply means you have an "inner" and an "outer" KroneckerSumMap. Given the nearly perfect coverage, one could try to rewrite things gently internally to flatten that structure, but this doesn't have high priority, I'd say.

src/conversion.jl
```diff
@@ -62,25 +64,29 @@ SparseArrays.sparse(A::WrappedMap) = sparse(A.lmap)
...
     mats = map(A->getfield(A, :lmap), maps)
     return convert(SparseMatrixCSC, sum(mats))
 end
 
 # CompositeMap
-function Base.Matrix{T}(AB::CompositeMap{<:Any,<:Tuple{MatrixMap,LinearMap}}) where {T}
+# Are all these methods below really useful?
+# The conversion to matrices is not something that belongs to ordinary use of LinearMaps.
+# It's rather for debugging purposes, where efficiency is not the primary concern.
+# Also, I don't think they are really that much more efficient
```
dkarrasch, line 88: Generally, the `Matrix` conversion is used in the Kronecker code in the vec trick, which is why I started introducing these methods (at least with the intention to write code that is faster than the columnwise mat-vec-muls):

https://github.com/Jutho/LinearMaps.jl/blob/b38c1ac1900d4751237bbef62927477017f4f7d0/src/kronecker.jl#L121-L131

`X` there is promoted to a `MatrixMap`, that's why we have all those `MatrixMap` conversion methods here.

src/conversion.jl
```diff
@@ -92,39 +98,48 @@ end
...
-function SparseArrays.sparse(AB::CompositeMap{<:Any,<:Tuple{MatrixMap,MatrixMap}})
+function SparseArrays.sparse(AB::CompositeMap{<:Any, <:Tuple{MatrixMap, MatrixMap}})
     B, A = AB.maps
     return convert(SparseMatrixCSC, A.lmap*B.lmap)
 end
-function Base.Matrix{T}(λA::CompositeMap{<:Any,<:Tuple{MatrixMap,UniformScalingMap}}) where {T}
+function Base.Matrix{T}(λA::CompositeMap{<:Any, <:Tuple{MatrixMap, UniformScalingMap}}) where {T}
     A, J = λA.maps
     return convert(Matrix{T}, J.λ*A.lmap)
+    # if you want less allocations: lmul!(J.λ, convert(Matrix{T}, A.lmap))
```
dkarrasch, line 118: I just tested this: it would change the underlying `A.lmap`, which is probably not what we want.

src/conversion.jl
```diff
@@ -92,39 +98,48 @@ end
...
     return convert(SparseMatrixCSC, A.lmap*J.λ)
+    # probably also works with sparse: rmul!(convert(SparseMatrixCSC, A.lmap), J.λ)
 end
 
 # BlockMap & BlockDiagonalMap
 Base.Matrix{T}(A::BlockMap) where {T} = hvcat(A.rows, convert.(Matrix{T}, A.maps)...)
 Base.convert(::Type{AbstractMatrix}, A::BlockMap) = hvcat(A.rows, convert.(AbstractMatrix, A.maps)...)
+# seems like code duplication:
+# convert(AbstractMatrix, A) will anyway fall back to Matrix{eltype(A)}(A)
+
```
dkarrasch, line 142: True. But conversion to `AbstractMatrix` keeps special matrix types, where `Matrix` makes them dense.

src/conversion.jl
Jutho: Ok I did not consider that.

src/conversion.jl
Jutho: Ok, now I see. I was tired last night :-).

src/conversion.jl
```diff
@@ -159,6 +178,9 @@ function Base.convert(::Type{AbstractMatrix}, L::KroneckerSumMap)
     IB = Diagonal(ones(Bool, size(B, 1)))
     return kron(convert(AbstractMatrix, A), IB) + kron(IA, convert(AbstractMatrix, B))
 end
+# seems like code duplication:
+# convert(AbstractMatrix, L) will anyway fall back to Matrix{eltype(L)}(L)
+
```
Jutho, line 183: Writing here, as I cannot comment on unchanged lines, but there seems to be an error then in line 188 (new file) or 166 (old line) below, where `convert(AbstractMatrix, A)` should probably be `sparse(A)` or `convert(SparseMatrixCSC, A)` and similar for `B`? Is there a reason to use `convert(SparseMatrixCSC, A)` rather than `sparse(A)` in some of the other sparse conversion methods above?

dkarrasch: Below, the end result is sparse anyway because `IA` and `IB` are sparse. That suffices to be handled by a `SparseArray.kron` method. Given that, I used the "weakest" conversion method possible, instead of enforcing sparsity on `A` and `B` (which might result in higher memory allocation, when these are `Diagonal`, `Bidiagonal` or stuff like that).

src/conversion.jl
dkarrasch: Don't worry. Sometimes I'm tired during the day. 😂

src/conversion.jl
Jutho: Ok, seems like a lot of considerations have gone into all of this which I fail to appreciate by a quick reading.

So maybe one final question: why do several of the methods above use `convert(SparseMatrixCSC, ...)` rather than just `sparse(...)`, e.g. lines  113, 122, 133, 146, 153, 164 ? It seems like defining the `sparse(...)` method is our entry point for this functionality, and we have lines 44 and 45 to handle the case where a user uses a different approach (namely `convert(SparseMatrixCSC, ...)` or just `SparseMatrixCSC(...)`) to do the sparse conversion. But then within the `sparse` definitions, we actually also use a different approach, namely `convert(SparseMatrixCSC, ...)`.

dkarrasch: I guess this is where you got me. You're right, the root function is `sparse`, which should be overloaded when needed. And by default, this will overload the other `convert` and `SparseMatrixCSC` methods. This is with regards to "the left hand side", the functions to be defined. With regards to the "right hand sides" (how we do it), I use `convert(SparseMatrixCSC, ...)` because it's argument `...` may happen to be sparse already, and in that case, I want to avoid creating a copy. Does that make sense?

Jutho: Ok, and `sparse(A::SparseMatrixCSC)` creates a copy, I see. I think this then settles it all. I will remove my comments and this will then probably be my final commit.

src/blockmap.jl
dkarrasch: Does this comment still exist?

src/scaledmap.jl
dkarrasch: This doesn't seem to be outdated, yet.

src/functionmap.jl
dkarrasch: Another decision to be made: what shall we do with `parent`? Just remove? I think it's used internally only in one spot in the kronecker code, and could be worked around with a ternary if-clause, as I had it originally.

src/composition.jl
dkarrasch: I guess this comment can also be removed, right?

Jutho:
Ok, I thought I already left a message after previous commit, but clearly not. So I think now all comments have been removed (thanks @dkarrasch for checking carefully). The only remaining issue, which I did not address, is how to deal with `parent`? I leave that up to you. I am fine with this or any further updates of this to become the new version 3.0. Thanks for perfect maintenance and improvement of this package for the last few years.

dkarrasch:
I chose to remove `parent` altogether. We need a clear general API first, and then we can go and implement it for the different types. It hasn't been released yet, so there is no harm removing it.

dkarrasch:
Thank you so much, Jutho! Now let's discuss the FillMap constructor over at #113, and we're done with v3.0. 😂

src/blockmap.jl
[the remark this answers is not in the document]
dkarrasch: This was a very helpful remark, Jutho! Overloading `Base._cat` instead of `Base.cat` allows to relax the requirement of "a LinearMap object among the first 8 arguments"! I'll include that in #114. 🎉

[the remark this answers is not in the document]
Jutho: Oh yes, I completely forgot about this, I was planning to change that myself, but if you can do this, way better :-)

Issue #125, opened 2021-01-05
performance issue with Kronecker product and complex values


vsaase:
Hi,
in this example I see a loss of efficiency for complex value:
```
using LinearAlgebra
using SparseArrays

spd = SparseMatrixCSC{ComplexF64}(I, 1000, 1000)
A = ones(2,2)
x = ones(2000)
B = kron(spd,A)
Bi = kron(spd, im*A)
Blm = kron(spd,LinearMap(A))
Blmi = kron(spd,im*LinearMap(A))

julia> @time B*x;
  0.000065 seconds (2 allocations: 31.328 KiB)
julia> @time Bi*x;
  0.000041 seconds (2 allocations: 31.328 KiB)
julia> @time Blm*x;
  0.000086 seconds (7 allocations: 47.266 KiB)
julia> @time Blmi*x;
  0.006984 seconds (1.01 k allocations: 187.969 KiB)

using Profile
@profile Blmi*x;
``` 
thanks for your work!,
Victor

dkarrasch:
Thanks @vsaase for the report and the interesting benchmark problems. First of all, the issue is not about the complex factor `im` by itself: a real number would have resulted in the same slowdown. The issue is that, as was, we took the Kronecker product of a `WrappedMap` and a `ScaledMap`. That seems to complicate things further down the road, so the solution is to hoist out the scaling factor in Kronecker products. Note that any manipulation in `LinearMaps.jl` is lazy, so matrices or scaling factors are always taken as they are, and not silently promoted to something that some heuristic might guess is beneficial. This lets the user have full control (aka responsibility 😉 ).

I guess that much of your code above is just for a quick description of the issue, but in case some aspects of it are related to your work, let me comment and suggest a few improvements.

1. There is no need to convert the identity `I` to a sparse matrix, and thereby occupy 3 vectors of length 1000. You can replace this one by `LinearMap(I, 1000)`.
2. For matrices filled with a single value like your `A`, we have the `FillMap` type. It only stores the fill value and the size, nothing else.
3. Your case is a little special in that your `x` vector is real, but the matrices are complex. Since kron-vec-multiplication is implemented via the vec-trick, we call `mul!` potentially several times. When you have different eltypes, you may end up with the `generic_mat_vec_mul!`, which is of course slower than BLAS. So you may need to play a bit with the eltypes before you start with real big problems.

I was working with the following modification of your code above.
```julia
[...]

PR #127, opened 2021-01-05
Restrict ScaledMap to RealOrComplex

dkarrasch:
While investigating #125, I also noticed that we should enforce commutative eltypes in `ScaledMap`s more strongly, as @JeffFessler probably had in his first draft, which later got spoiled by me. Otherwise the results may be wrong when you have a quaternion-map and/or -vector and a complex scaling. I'll probably need to add a test case later.


dkarrasch:
I'd merge this soon, if nobody objects.

PR #131, opened 2021-01-13
Add in-place left-multiplication for absmat

### LINKED ISSUE #129: Missing method mul!(::Matrix,::Matrix,::LinearMap)
Hi,

It seems like the method for in-place multiplication where the final argument is a `LinearMap` is missing:
```julia
using LinearMaps, LinearAlgebra
N = 10
l = LinearMap(identity, N)
A = rand(N,N)

B = similar(A)
mul!(B, l, A) # works in this order
mul!(B, A, l) # MethodError
```

I'm on Julia v1.5.3 and LinearMaps v3.1.0.

dkarrasch:
Closes #129.

### DIFF
--- a/docs/src/types.md
+++ b/docs/src/types.md
@@ -102,6 +102,7 @@ Base.:*(::LinearMap,::AbstractMatrix)
 Base.:*(::AbstractMatrix,::LinearMap)
 LinearAlgebra.mul!(::AbstractVecOrMat,::LinearMap,::AbstractVector)
 LinearAlgebra.mul!(::AbstractVecOrMat,::LinearMap,::AbstractVector,::Number,::Number)
+LinearAlgebra.mul!(::AbstractMatrix,::AbstractMatrix,::LinearMap)
 *(::LinearAlgebra.AdjointAbsVec,::LinearMap)
 *(::LinearAlgebra.TransposeAbsVec,::LinearMap)
 ```
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -231,17 +231,6 @@ function _generic_mapmat_mul!(Y, A, X, α=true, β=false)
     return Y
 end
 
-function mul!(Y::AbstractMatrix, X::AbstractMatrix, A::LinearMap)
-    check_dim_mul(Y, X, A)
-    mul!(Y', A', X')
-    return Y
-end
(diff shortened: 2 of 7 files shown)


dkarrasch:
@JeffFessler Do you have any additional comments here? This might be somewhat related to your interests/use cases.

JeffFessler:
This definitely is a valuable addition.  I would have guessed `Y::AbstractVecOrMat` might have allowed to cover the vector and matrix case in a unified way but I have not thought about it carefully and I am sure you would have done that if it were possible!  I just took a quick skim and it looked fine.  Thanks!

dkarrasch:
Well, that's at least worth a thought (about `AbstractVecOrMat`), but we do have the convention (consistent with Base), that input vector can yield output VecOrMat, but input matrix can only give an output matrix, even though it may fit into a vector:
```julia
julia> mul!(zeros(3), rand(3, 3), rand(3, 1))
ERROR: MethodError: no method matching mul!(::Array{Float64,1}, ::Array{Float64,2}, ::Array{Float64,2}, ::Bool, ::Bool)
Closest candidates are:
  mul!(::AbstractArray, ::Number, ::AbstractArray, ::Number, ::Number) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/generic.jl:124
  mul!(::AbstractArray, ::AbstractArray, ::Number, ::Number, ::Number) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/LinearAlgebra/src/generic.jl:126
  mul!(::StridedArray{T, 1}, ::StridedVecOrMat{T}, ::StridedArray{T, 1}, ::Number, ::Number) where T<:Union{Complex{Float32}, Complex{Float64}, Float32, Float64} 
```
Considering a LinearMap as equivalent to a matrix, then I think the output should be a matrix.

Okay, so I think I'll merge this and make one more person (even more) satisfied with `LinearMaps.jl`! 😆

dkarrasch:
Boy, this whole non-commutative stuff is driving me crazy. I'll need to add some tests for it, and then merge.

JeffFessler:
ah of course.  dumb question.    as a reminder to myself, this "left" multiply works:
`mul!(zeros(1,4), ones(1,3), ones(3,4))`
but this does not, understandably:
`mul!(zeros(4), ones(1,3), ones(3,4))`

PR #136, opened 2021-02-24
Allow wrapping of vectors, extend Kronecker product

dkarrasch:
This treats vectors upon wrapping in `LinearMap` like a column matrix, without reshaping however. This allows to easily represent rank-1 operators as `v ⊗ w'` with optimal performance.


dkarrasch:
Are there any objections against starting to treat vectors as column matrices in `LinearMaps` constructors? If not, I think this is a nice generalization, and there was a request for that in the Slack channel for lazy outer products.

Jutho:
I guess that is indeed fine; I see no obstructions.

PR #137, opened 2021-03-04
Fix performance issue when RHS is a Matrix

rasmushenningsson:
Before this patch, `mul!` would incorrectly call `_generic_mapmat_mul!` instead of `_unsafe_mul!` in cases when the RHS was a matrix.

This problem lead to huge performance penalties in all cases where the LinearMap acts more efficiently on a matrix than on the individual columns.

```
using LinearMaps
using BenchmarkTools
using LinearAlgebra
A = randn(10000,1000);
AL = LinearMap(A);
B = randn(1000,1000);
C = zeros(size(A,1), size(B,2));
```

Before patch:
```
julia> @btime mul!($C,$A,$B);
  139.193 ms (0 allocations: 0 bytes)

julia> @btime mul!($C,$AL,$B);
  2.812 s (0 allocations: 0 bytes)
```

After patch:
```
julia> @btime mul!($C,$A,$B);
  142.395 ms (0 allocations: 0 bytes)

julia> @btime mul!($C,$AL,$B);
  147.012 ms (0 allocations: 0 bytes)
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LinearMaps"
 uuid = "7a12625a-238d-50fd-b39a-03d52299707e"
-version = "3.2.1"
+version = "3.2.2"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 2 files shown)


dkarrasch:
Awesome! Nice catch. That logic was a bit tricky, and I see that we took the turn to the generic function one step too early. Could you bump the patch version number in Project.toml, then we merge and release. Thank you very much!

rasmushenningsson:
Should be good to go! :)
Thanks for a nice package and a quick response!

PR #143, opened 2021-03-15
Fix mul!(y, A::CompositeMap, x::AbstractMatrix)

rasmushenningsson:
Performance improvement by avoiding fallback to _generic_mapmat_mul! for
mul!(y, A::CompositeMap, x::AbstractMatrix).

Benchmark:
```
using LinearMaps
using LinearAlgebra
using BenchmarkTools
N = 1000;
A = LinearMaps.CompositeMap{Float64}((LinearMap(randn(N,N)),));
B = randn(N,N);
C = randn(N,N);
X = randn(N,N);
Y = zeros(N,N);
AB = A*B;
ABC = AB*C;

# Before patch
julia> @btime mul!($Y,$A,$X);
  74.417 ms (0 allocations: 0 bytes)

julia> @btime mul!($Y,$AB,$X);
  293.908 ms (1000 allocations: 7.75 MiB)

julia> @btime mul!($Y,$ABC,$X);
  626.289 ms (2000 allocations: 15.50 MiB)


# After patch
julia> @btime mul!($Y,$A,$X);
  11.512 ms (0 allocations: 0 bytes)

julia> @btime mul!($Y,$AB,$X);
  26.461 ms (2 allocations: 7.63 MiB)

julia> @btime mul!($Y,$ABC,$X);
  39.346 ms (4 allocations: 15.26 MiB)
```

Note that I change the source/dest allocation strategy for `CompositeMap`s with >= 3 maps.
Previously, a `resize!` was performed in a try-catch. I think there is no general benefit of a `resize!` here, since it might copy the data if resizing to a bigger array. Furthermore, `resize!` is clumsy to use in the matrix case.
So, I changed to a simpler strategy of only reusing the array if it's exactly the same size.

If I've missed some benefit of using `resize!` I can revert to the previous behavior and update the pull request.

### DIFF
--- a/src/composition.jl
+++ b/src/composition.jl
@@ -129,19 +129,23 @@ LinearAlgebra.adjoint(A::CompositeMap{T}) where {T} =
 # comparison of CompositeMap objects
 Base.:(==)(A::CompositeMap, B::CompositeMap) = (eltype(A) == eltype(B) && A.maps == B.maps)
 
-# multiplication with vectors
+# multiplication with vectors/matrices
 _unsafe_mul!(y::AbstractVecOrMat, A::CompositeMap, x::AbstractVector) =
     _compositemul!(y, A, x)
+_unsafe_mul!(y::AbstractMatrix, A::CompositeMap, x::AbstractMatrix) =
+    _compositemul!(y, A, x)
+
 function _compositemul!(y::AbstractVecOrMat,
                         A::CompositeMap{<:Any,<:Tuple{LinearMap}},
-                        x::AbstractVector,
+                        x::AbstractVecOrMat,
                         source = nothing,
                         dest = nothing)
     return _unsafe_mul!(y, A.maps[1], x)
 end
 function _compositemul!(y::AbstractVecOrMat,
-                        A::CompositeMap{<:Any,<:Tuple{LinearMap,LinearMap}}, x::AbstractVector,
-                        source = similar(y, size(A.maps[1], 1)),
+                        A::CompositeMap{<:Any,<:Tuple{LinearMap,LinearMap}},
+                        x::AbstractVecOrMat,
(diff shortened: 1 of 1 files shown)


dkarrasch:
I don't think `resize!` copies data, but I'm not sure. It does some `ccall` eventually, which is... above my paygrade. To play it safe, how about having a vector and a matrix version of the multi-map case?

rasmushenningsson:
`resize!` does (often) copy data if you change to a larger size vector, because it promises that the existing elements will be kept, and it's unlikely that the array can grow in place. So a new block of memory has to be allocated, which means the data has to be copied to that block.
You can see that the underlying pointer changes below when we `resize!` to a larger array, but not to a smaller.

```
julia> x = zeros(2);

julia> pointer(x)
Ptr{Float64} @0x00007fed2758f6b0

julia> resize!(x,4)
4-element Array{Float64,1}:
 0.0
 0.0
 9.6e-322
 8.55e-322

julia> pointer(x)
Ptr{Float64} @0x00007fed27846a70

julia> resize!(x,2);

julia> pointer(x)
Ptr{Float64} @0x00007fed27846a70
```

But I've updated the code to keep the old solution for the vector case. No reason to change it really - unless I want to spend some time benchmarking different scenarios. :)

dkarrasch:
That's interesting. I thought it wouldn't because
```julia
y = x = zeros(2)
resize!(x, 5)
y === x # true
```
but I'm not familiar enough with the details to tell what's really going on under the hood.

dkarrasch:
For the matrix case, I believe this is reasonable enough. In our previous `_generic_mapmat_mul!` branch, we would allocate the intermediate vectors for each column to which we apply the `CompositeMap`, so it ends up allocating basically as much.

Thanks @rasmushenningsson for another nice contribution.

PR #145, opened 2021-04-14
Add IndexableMap

### LINKED ISSUE #38: Why no getindex() ?
Why is `getindex` not supported (quite explicitly in the README) for LinearMaps?

It seems like it would be pretty easy to provide the functionality as a convenience to users even if it would be somewhat inefficient.  For example here is a rough start on `A[i,j]`

```
function getindex(A::LinearMap, i::Int, j::Int) # for A[i,j]
  e = zeros(size(A,2)); e[j] = 1
  tmp = A * e
  return tmp[i]
end
```

I am less sure how getindex works for `A[:,j]` which is what I really want.
Of course I could do this type of overloading myself locally, but why not benefit all users?
I've done this in Matlab by the way so I expected it to be even easier in Julia!

dkarrasch:
This adds a type for indexable LinearMaps. Construction is via a keyword argument `getind`, naming can be discussed, of course.

Closes #38.


src/LinearMaps.jl
```diff
@@ -293,7 +294,9 @@ For the function-based constructor, there is one more keyword argument:
     The default value is guessed by looking at the number of arguments of the first
     occurrence of `f` in the method table.
 """
-LinearMap(A::MapOrVecOrMat; kwargs...) = WrappedMap(A; kwargs...)
+LinearMap(A::MapOrVecOrMat; getind=nothing, kwargs...) = _LinearMap(getind, A; kwargs...)
```
JeffFessler, line 297:
```suggestion
LinearMap(A::MapOrVecOrMat; getind=Base.getindex, kwargs...) = _LinearMap(getind, A; kwargs...)
```
For a `WrappedMap` could the default `getind` simply be `Base.getindex` because presumably that is the expected behavior for a matrix.

src/indexablemap.jl
```diff
@@ -0,0 +1,49 @@
...
+LinearAlgebra.ishermitian(A::IndexableMap) = ishermitian(A.lmap)
+LinearAlgebra.isposdef(A::IndexableMap) = isposdef(A.lmap)
+
+Base.:(==)(A::IndexableMap, B::IndexableMap) = A.lmap == B.lmap
+
+Base.adjoint(A::IndexableMap) = IndexableMap(adjoint(A.lmap), (i,j) -> adjoint(A.getind(j,i)))
+Base.transpose(A::IndexableMap) = IndexableMap(transpose(A.lmap), (i,j) -> transpose(A.getind(j,i)))
+# rewrapping preserves indexability but redefines, e.g., symmetry properties
+LinearMap(A::IndexableMap; getind=nothing, kwargs...) = IndexableMap(LinearMap(A.lmap; kwargs...), getind)
+# addition/subtraction/scalar multiplication preserve indexability
```
JeffFessler, line 19: nice!
I wish I had a suggestion for how to handle a `CompositeMap`...

Jutho:
I won't have time to review this before next week, but I just thought of another possible approach. We could support a general `getindex` keyword argument in `FunctionMap`, the default value of which is a function that evaluates the multiplication with a unit vector with a single entry one, but where this default function refuses to work depending on the value of a global switch, similar to how CUDA.jl has `allowscalar` . The user then has two options to have support for `getindex`, either by providing a `getindex` method upon construction of the `LinearMap`, or by setting `allowscalar(true)`, knowing that this might lead to slow/suboptimal code. Not sure what is needed to set this up, but this might be useful and does not require the introduction of new LinearMap subtypes.

src/uniformscalingmap.jl
```diff
@@ -6,12 +6,16 @@ struct UniformScalingMap{T} <: LinearMap{T}
...
+@deprecate UniformScalingMap(λ::Number, M::Int, N::Int) UniformScalingMap(λ, M) false
+@deprecate UniformScalingMap(λ::Number, sz::Dims{2}) UniformScalingMap(λ, sz[1]) false
+# the following methods are misleading: they introduce the opportunity to pass 2 dims,
+# but throw if the dims are unequal!
+# UniformScalingMap(λ::Number, M::Int, N::Int) =
+#     (M == N ?
+#         UniformScalingMap(λ, M) : error("UniformScalingMap needs to be square"))
+# UniformScalingMap(λ::T, sz::Dims{2}) where {T} =
+#     (sz[1] == sz[2] ?
+#         UniformScalingMap(λ, sz[1]) : error("UniformScalingMap needs to be square"))
```
JeffFessler, line 19:
```suggestion
Base.getindex(A::UniformScalingMap, args...) = A.λ * A.lmap.getind(args...)
```
Seems like something like this should be the natural default for this type.

JeffFessler:
I was going to suggest something similar to what @Jutho said, i.e., some way of letting users "opt in" to the natural default.  It could be a switch or an exported convenience routine that could be passed back in like `getind=getind_unitvector_default`.  The switch seems easier...

I put in a couple other suggestions.

Issue #149, opened 2021-06-12
isposdef error for composite maps


JeffFessler:
The recent revelation (to me) in #147 that `isposdef` might be called when forming a wrapped map made me want to look at that more carefully because I use a lot of combinations of LMs and want to make sure that such combinations are performant.  In the process of trying to construct a timing test, I found the following bug:

```julia
using LinearMaps
using LinearAlgebra: isposdef

A = zeros(2)
L = LinearMap(A)
H = L * L'
@show isposdef(H), isposdef(Matrix(H)), isposdef(A*A') # oh no!
```

Here `H` is akin to 2x2 matrix of 0s yet it is being reported incorrectly as positive definite.

I tried to diagnose so I could make a PR but `@eval` code is a bit hard for me to parse.  My hunch is that the problem is somewhere near here:
https://github.com/Jutho/LinearMaps.jl/blob/032832e7d887bf9ad94e19ade4aa6c2044a31039/src/composition.jl#L31

The product `A * A'` is alway positive semidefinite, but it is not positive definite in general, unless `A` has full row rank (which would be expensive to verify).
So that symmetry rule is insufficient for ensuring positive definiteness.
I recommend separating the `isposdef` rules from the symmetry rules.
`A * B * C * B * A` and `A * B * C * B' * A'` and `A * B' * C * B * A` etc. are all positive definite *if* (sufficient condition) A,B,C are all positive definite.  (I'm using the usual convention that Hermitian symmetry is assumed as part of positive definiteness.)

dkarrasch:
This is a duplicate of #148, but I'll keep this one as it has more text and a concrete proposal. 🤣

dkarrasch:
Fixed by https://github.com/Jutho/LinearMaps.jl/pull/151.

PR #151, opened 2021-06-12
isposdef for compositemap

JeffFessler:
attempt to address #149


JeffFessler:
ok, now it finally passes the tests because i removed some tests.
before worrying about code coverage, let's see what you think about the issue and possible solution...

src/composition.jl
```diff
@@ -42,6 +41,16 @@ for (f, _f, g) in ((:issymmetric, :_issymmetric, :transpose),
...
 end
 
+# A * B * A and A * B * A' are positive definite if (sufficient condition) A & B are positive definite
+LinearAlgebra.isposdef(A::CompositeMap) = _isposdef(A.maps)
+_isposdef(maps::Tuple{}) = true # empty product is equivalent to "I" which is pos. def.
+_isposdef(maps::Tuple{<:LinearMap}) = isposdef(maps[1])
+function _isposdef(maps::Tuple{Vararg{LinearMap}})
+    (maps[end] == adjoint(maps[1]) || maps[end] == maps[1]) && 
+    isposdef(maps[1]) && isposdef(maps[end]) &&
+    _isposdef(Base.front(Base.tail(maps)))
```
dkarrasch, line 51: I believe it suffices to check one of the foregoing conditions are met?
```suggestion
    isposdef(maps[1]) && _isposdef(Base.front(Base.tail(maps)))
```

test/scaledmap.jl
```diff
@@ -77,11 +77,11 @@ using Test, LinearMaps, LinearAlgebra, BenchmarkTools
...
 
     # check scale*conj(scale)
-    A = LinearMap{Float32}(rand(N, 2)) # rank=2 w.p.1
+    A = LinearMap{Float32}(rand(1, 2)) # rank=1 w.p.1
     B = @inferred β * A
     C = @inferred B' * B
-    @test @inferred isposdef(C)
-
+    @test !isposdef(C)
+
```
dkarrasch, line 84:
```suggestion

```

test/composition.jl
```diff
@@ -27,8 +27,8 @@ using Test, LinearMaps, LinearAlgebra, SparseArrays
     @test length(F4.maps) == 4
     @test @inferred F4 * v == @inferred F * (F * (F * (F * v)))
     @test @inferred Matrix(M * transpose(M)) ≈ A * transpose(A)
-    @test @inferred !isposdef(M * transpose(M))
-    @test @inferred isposdef(M * M')
+#   @test @inferred !isposdef(M * transpose(M))
+#   @test @inferred isposdef(M * M')
```
dkarrasch, line 31:
```suggestion
    @test @inferred !isposdef(M * transpose(M))
    @test @inferred isposdef(LinearMap(M * M', isposdef=true))
```

test/composition.jl
```diff
@@ -40,10 +40,10 @@ using Test, LinearMaps, LinearAlgebra, SparseArrays
...
     @test @inferred ishermitian(FC'FC)
     @test @inferred ishermitian(FC'*H*FC)
-    @test @inferred isposdef(transpose(F) * F * 3)
-    @test @inferred isposdef(transpose(F) * 3 * F)
-    @test @inferred !isposdef(-5*transpose(F) * F)
-    @test @inferred isposdef((M * F)' * M * 4 * F)
+#   @test @inferred isposdef(transpose(F) * F * 3)
+#   @test @inferred isposdef(transpose(F) * 3 * F)
+#   @test @inferred !isposdef(-5*transpose(F) * F)
+#   @test @inferred isposdef((M * F)' * M * 4 * F)
```
dkarrasch, line 46: Let's change this to `issymmetric` to check for the destructuring logic:
```suggestion
    @test @inferred issymmetric(transpose(F) * F * 3)
    @test @inferred issymmetric(transpose(F) * 3 * F)
    @test @inferred !isposdef(-5*transpose(F) * F)
    @test @inferred ishermitian((M * F)' * M * 4 * F)
```

dkarrasch (commented):
Some ideas to change the commented tests to something useful. The main question remains whether we want to keep some `isposdef`-logic at all or not (for `CompositeMaps`).

src/composition.jl
```diff
@@ -42,6 +41,15 @@ for (f, _f, g) in ((:issymmetric, :_issymmetric, :transpose),
     end
 end
 
+# A * B * A and A * B * A' are positive definite if (sufficient condition) A & B are positive definite
+LinearAlgebra.isposdef(A::CompositeMap) = _isposdef(A.maps)
```
dkarrasch, line 45: With the latest #154, this should be
```suggestion
_isposdef(A::CompositeMap) = _isposdef(A.maps)
```
if I'm not mistaken. Since we are very conservative about checking positive definiteness, I think we should give `CompositeMaps` a chance, in case somebody bothered to manually set the `isposdef` flag `true` for the factors. I mean I'm in favor of keeping this sufficient criterion instead of just defaulting to `false` for any `CompositeMap`.

JeffFessler: I agree that we should see if the sufficient condition holds.
I tried to sync my branch with main branch but somehow my attempt to resolve the conflict didn't work.
I hope I didn't mess it up...

test/composition.jl
```diff
@@ -132,4 +132,10 @@ using Test, LinearMaps, LinearAlgebra, SparseArrays
     X = rand(size(Ls, 2), 10)
     Y = similar(X, (size(Ls, 1), size(X, 2)))
     @test mul!(Y, Ls, X) ≈ L4.lmap * L3.lmap * L2.lmap * L1.lmap * X
+
+    # test isposdef on a case where sufficient conditions work
+    B = LinearMap([1 0; 0 1]) # isposdef!
```
dkarrasch, line 137: The tests in lines 139-140 fail, because `B` itself is not recognized as positive definite. Those two should be fixed with
```suggestion
    B = LinearMap([1 0; 0 1], isposdef=true) # isposdef!
```

src/composition.jl
dkarrasch: Sorry, I confused myself. This was correct. `_isposdef(C::CompositeMap)` will fall back to `isposdef(C)` which will then enter the recursion here.

dkarrasch:
After the revision of my wrong proposal, I think you could directly bump the minor version number and then we merge and release.

src/composition.jl
```diff
@@ -42,6 +41,15 @@ for (f, _f, g) in ((:issymmetric, :_issymmetric, :transpose),
     end
 end
 
+# A * B * A and A * B * A' are positive definite if (sufficient condition) A & B are positive definite
+_isposdef(A::CompositeMap) = _isposdef(A.maps)
```
dkarrasch, line 45:
```suggestion
LinearAlgebra.isposdef(A::CompositeMap) = _isposdef(A.maps)
```

JeffFessler:
seems ok now.  it is curious to me that nightly is failing due to method ambiguities, but maybe save that for another day :)

dkarrasch:
Yes, the method ambiguity is already taken care of #158.

PR #154, opened 2021-06-14
Speed up `WrappedMap` constructor and LM-promotion

### LINKED ISSUE #152: Positive definiteness of wrapped maps and sums
Related to #148, from a performance point of view it would be preferable to default to `false` for a sum of LMs as well in my view.  Here is a MWE that is fairly close to actual use cases.

```julia
using LinearMaps
using BenchmarkTools
using SparseArrays

N = 2^16
IN = I(N)
S = sparse(IN)
L = LinearMap(x -> sum(x), y -> repeat(y, N), 1, N) # toy example
C = LinearMap(x -> circshift(x,-1) - x, y -> circshift(y,1) - y, N, N) # realistic regularizer
Hl = L' * L
Hc = C' * C

function fi(Hl, Hs, IN)
    Hl + Hc + IN # Hessian matrix with additional Tikhonov regularizer
end

function fs(Hl, Hs, S)
    Hl + Hc + S
end

@btime fi($Hl, $Hc, $IN) # 20.437 μs (20 allocations: 1.38 KiB)
@btime fs($Hl, $Hc, $S) #  6.501 ms (61 allocations: 14.50 MiB)
```

I realize that in this simple case one should just use `I` rather than a sparse matrix, but I'm pretty sure there are other examples that would be less obvious.

At first I thought we should replace the `all` below with `false`.

https://github.com/Jutho/LinearMaps.jl/blob/032832e7d887bf9ad94e19ade4aa6c2044a31039/src/linearcombination.jl#L24

But then I realized that the actual culprit here is probably the default for a wrapmap:
https://github.com/Jutho/LinearMaps.jl/blob/032832e7d887bf9ad94e19ade4aa6c2044a31039/src/wrappedmap.jl#L10

My preference would be to make that `false` if the input is an AbstractMatrix.
(I'd do the same for the symmetry conditions too, because I suspect those could still be expensive for a large sparse non-diagonal matrix.)

(If wrapmap defaults to false then the `all` for a sum is fine.)
This time I'm not submitting a PR because this would be a bigger breaking change...

dkarrasch:
I agree with @JeffFessler that extensive property checks go somewhat against the idea of lazy XYZ, so one option would be to redirect the (symmetry/isposdef) checks to internal functions which yield `true` in very cheap cases and otherwise err towards `false`. I think that is reasonable since, for instance, when people work with `Matrix`es, they don't check for `isposdef` either, typically. Same for symmetry. Now if people really care about these properties, they likely wrap their matrices in related types, like `Symmetric`, `Hermitian`, or `PDMats`. Other library authors could then overload our internal `_is*` functions. Opinions?

Closes #152.

### DIFF
--- a/docs/src/history.md
+++ b/docs/src/history.md
@@ -1,5 +1,15 @@
 # Version history
 
+## What's new in v3.4
+
+* In `WrappedMap` constructors, as implicitly called in addition and mutliplication
+  of `LinearMap`s and `AbstractMatrix` objects, (conjugate) symmetry and positive
+  definiteness are only determined for matrix types for which these checks are expected
+  to be very cheap or even known at compile time based on the concrete type. The default
+  for `LinearMap` subtypes is to call, for instance, `issymmetric`, because symmetry
+  properties are either stored or easily obtained from constituting maps. For custom matrix
+  types, define corresponding methods `LinearMaps._issymmetric`, `LinearMaps._ishermitian` and `LinearMaps._isposdef` to hook into the property checking mechanism.
+
 ## What's new in v3.3
 
 * `AbstractVector`s can now be wrapped by a `LinearMap` just like `AbstractMatrix``
--- a/docs/src/history.md
+++ b/docs/src/history.md
@@ -8,7 +8,8 @@
   to be very cheap or even known at compile time based on the concrete type. The default
   for `LinearMap` subtypes is to call, for instance, `issymmetric`, because symmetry
   properties are either stored or easily obtained from constituting maps. For custom matrix
-  types, define corresponding methods `LinearMaps._issymmetric`, `LinearMaps._ishermitian` and `LinearMaps._isposdef` to hook into the property checking mechanism.
+  types, define corresponding methods `LinearMaps._issymmetric`, `LinearMaps._ishermitian`
+  and `LinearMaps._isposdef` to hook into the property checking mechanism.
 
 ## What's new in v3.3
 
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -48,7 +48,7 @@ LinearAlgebra.isposdef(::LinearMap) = false # default assumptions
 
 Base.ndims(::LinearMap) = 2
 Base.size(A::LinearMap, n) =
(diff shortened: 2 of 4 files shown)


dkarrasch:
This is breaking, of course, and should go together with a possible change in the check for `isposdef(::CompositeMap)`, be it removal or strengthening.

JeffFessler:
I like the (optional) overload of `_is*` idea.

I was thinking about this more and there are quite a few other matrix traits that users might care about in selected situations (invertible, circulant, tridiagonal, sparse, block structure, ...).  I'm not suggesting that we add fields and try to track those, but rather than positive definiteness is just one of a long list and may not deserve special status if it comes at a performance penalty.  The `_is*` idea looks like it nicely addresses that.

dkarrasch:
@Jutho This is touching very old defaults. What do you think?

dkarrasch:
This also addresses #152, right?

JeffFessler:
yes, because of the default `_issymmetric(A::AbstractMatrix) = false`

Jutho:
I definitely agree that `isposdef` should not be applied by default when wrapping a matrix. This could be a very large (sparse) matrix for which `isposdef` is impossible to compute efficiently. `issymmetric/ishermitian` is typically not that expensive, and also works (more or less) efficiently for sparse matrices, but I am certainly fine with defaulting this to false, and letting the user handling this.

dkarrasch:
How about performing the check for symmetry/Hermitianness of sparse matrices? Or shall we let users define `_issymmetric(A::SparseMatrixCSC)` as `false` or `issymmetric(A)` as they wish? That would be type piracy, strictly speaking, right? The check should touch only the nonzero values (and their symmetric counterpart), so typically should not be too expensive, especially for generic sparse matrices, so I'm inclined to include sparse ones in the to-be-checked list.

JeffFessler:
I'm fine with your suggestion.  (My sparse use cases are generally non-square anyway.)
Updating the docstring for `LinearMap` should suffice to alert users about this.

BTW, right now it reads:
`For existing linear maps or matrices A, the default values will be taken by
  calling issymmetric, ishermitian and isposdef on the existing object A.`
Probably should update as part of this PR.

JeffFessler:
LGTM

PR #161, opened 2021-09-13
Check output size for FunctionMaps

### LINKED ISSUE #160: No warning for invalid sizes in FunctionMaps
A student of mine was debugging code and discovered that `FunctionMaps` do not report inconsistencies between the declared map size and the output of the function:

```julia
using LinearMaps
up = 2
x = zeros(2, 3)
m, n = size(x)
A = LinearMap(x -> zeros(eltype(x), up .* size(x)), up^2*m*n, m*n) # subtle bug here in output size
y = A * vec(x) # no error or warning!
```

Maybe this is a feature? 😄 

It could be addressed by adding a check after this line (and in similar places):

https://github.com/Jutho/LinearMaps.jl/blob/744800af0f07033f61112ad4213ac4ba7116bfaa/src/functionmap.jl#L67

like
```
size(y) == size(A,1) || throw(DimensionMismatch())
```

I guess the question is whether the "overhead" that all users would incur for the size check is worth it for saving some debugging time by other users.  Personally I'd be in favor of adding the check, and am willing to make the PR.  After all, we already check the input size:

https://github.com/Jutho/LinearMaps.jl/blob/744800af0f07033f61112ad4213ac4ba7116bfaa/src/functionmap.jl#L52

so it seems natural to check the output too.  The "pro" users who care about efficiency are more likely to use the mutating versions `mul!` that appropriately seem to lack size checks if I am reading `_unsafe_mul!` correctly.

Thoughts?

JeffFessler:
Closes #160.

Hopefully I found all the (non-mutating) cases where this check is relevant.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LinearMaps"
 uuid = "7a12625a-238d-50fd-b39a-03d52299707e"
-version = "3.4.0"
+version = "3.4.1"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
--- a/src/functionmap.jl
+++ b/src/functionmap.jl
@@ -43,6 +43,7 @@ function Base.:(*)(A::FunctionMap, x::AbstractVector)
         A.f(y, x)
     else
         y = A.f(x)
+        length(y) == A.M || throw(DimensionMismatch())
     end
     return y
 end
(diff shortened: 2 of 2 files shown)


dkarrasch (commented):
Cool. Seems to be complete. Can we have it be `size(A, 1)` consistently?

JeffFessler:
Sure!  The funny thing is that there was a A.N earlier in that function so I was trying to be consistent with that, but now I see that the various functions were sometimes using A.N and A.M and sometimes using size.
Now I think I've made them all consistent.

dkarrasch:
This looks good. I rerun the tests to see if some issue on master got fixed meanwhile. @JeffFessler Could you please bump the patch version (I consider this a bugfix)?

JeffFessler:
the nightly fail is due to some type inference problem it seems

dkarrasch:
> the nightly fail is due to some type inference problem it seems

Yes, I could reproduce that locally. It seems there is a magic number of arguments when it fails, like, the test two lines above the failing one (an `hcat` with three less map arguments) doesn't fail. Perhaps we need an issue upstream.

PR #162, opened 2021-10-06
Overload axes for ScaledMap and LinearCombination

krcools:
This allows for the linear combination to inherit the potentially rich structure of the terms (e.g. when the terms are BlockArrays)


krcools:
Thanks for pointing out the BlockMap type. At first sight there is some overlap in functionality but the two are not interchangeable for my specific application.

I'll look into writing some tests. Do I get it right that you want the tests to live in e.g. LiftedMaps.jl which then would become a test-only dependency of LinearMaps?

What you describe definitely is in line with the julia docs recommendations for supporting generalised Arrays:

https://docs.julialang.org/en/v1/devdocs/offset-arrays/#Generalizing-existing-code

krcools:
Sure. I am working on getting the tests to pass on earlier julia versions.

Can't seem to go back as far as julia-v.1.0 though. The BlockArrays releases compatible with julia-v1.0 (up to BlockArrays@0.10.2) do not yet support the axes API.

Shall I make the tests conditional to the julia version or do you suggest another approach?

dkarrasch:
I have seen first packages that put a lower bound of v1.6 on Julia, since that is (likely) going to be the next LTS. I believe at some point we may follow. But, the actual functionality is available in `Base` of Julia v1.0, right? So this is just a test issue.
I think it would be fine to have the tests run only on v1.6+, and therefore use the more "modern" functionality right away.

> As is, the user can define linear combinations with inconsistent axes. The reason is that axes equality as defined in BlockArrays returns true for all axes of equal length, even if they have different blocksizes.

You seem to have changed the `size` checks to `axes` checks in the `LinearCombination` constructor. Does that mean that you are stricter than `BlockArrays`?

test/nontradaxes.jl
```diff
@@ -0,0 +1,44 @@
...
+    @test axes(Nu)[1] == axes(N)[1] == axes(B)[1]
+    @test blocklengths(axes(Nu)[1]) == blocklengths(axes(N)[1]) == blocklengths(axes(B)[1]) == [2,3]
+
+    C = B + 2N
+    @test axes(C) === axes(B) === axes(N)
+    @test C*u ≈ 3*Nu
+
+    Cu = C*u
+    @test axes(C)[1] == ax1
+    @test blocklengths(axes(C)[1]) == blocklengths(ax1)
```
dkarrasch, line 43: Can you add another quick test for a `ScaledMap`, please? Just to have it covered.

dkarrasch (commented):
One last tiny bit, and then you can bump the minor version and add a little section in the `history.md` in the docs folder, so people can find a notice about the generalized `axes` feature. Then this should be good to go.

test/nontradaxes.jl
dkarrasch: Okay, I guess it's kind of covered by the `C = B + 2N` line.

dkarrasch (approved):

PR #163, opened 2021-11-02
Julia v1.6+ overhaul

dkarrasch:
~~This additional annotation seems necessary on current master. Otherwise it fails on map tuples longer than some number between 10 (which passes) and 13 (which fails), see our most recent commits.~~

EDIT: This PR started because one test in the block maps test suite started to fail on nightly/v1.8. It turned out that some functions that are called in the process of constructing `BlockMap`s had type inference difficulties. Apparently, quite some work was done even in the v1.6-v1.8 period regarding type inference in Julia Base, and due to changes pre-v1.6 it seemed impossible to maintain clear code that would equally run in a type-stable fashion on all Julia versions starting from v1.0. Given that v1.6 is gonna be the new LTS once v1.7 is released, I ended up removing all the `VERSION` branches and switched to by now well-established syntax. Lifting the lower compat bound to v1.6 shouldn't do much harm, as we are leaving a fully operational version behind, just in case some downstream code really requires Julia v1.0.

There is no rush to merge this, and we can leave it open for a while and see if we want to merge other urgent features before we lift the lower compat bound.

### DIFF
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -406,14 +406,8 @@ struct BlockDiagonalMap{T,
             promote_type(T, TA) == T ||
                 error("eltype $TA cannot be promoted to $T in BlockDiagonalMap constructor")
         end
-        # row ranges
-        inds = vcat(1, size.(maps, 1)...)
-        cumsum!(inds, inds)
-        rowranges = ntuple(i -> inds[i]:inds[i+1]-1, Val(length(maps)))
-        # column ranges
-        inds[2:end] .= size.(maps, 2)
-        cumsum!(inds, inds)
-        colranges = ntuple(i -> inds[i]:inds[i+1]-1, Val(length(maps)))
+        rowranges = _getranges(maps, 1)
+        colranges = _getranges(maps, 2)
         return new{T, As, typeof(rowranges)}(maps, rowranges, colranges)
     end
 end
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -24,6 +24,13 @@ BlockMap{T}(maps::As, rows::Rs) where {T, As<:LinearMapTuple, Rs} =
 
 MulStyle(A::BlockMap) = MulStyle(A.maps...)
 
+function _getranges(maps, dim, inds::NTuple{N,Int}=ntuple(identity, Val(length(maps)))) where {N}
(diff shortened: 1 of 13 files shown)


dkarrasch:
I wonder if there's something to be investigated here. It has to do with the length of the map tuple, and it's not quite clear where inference breaks down, or which threshold could be responsible for that.

Jutho:
I have not been following Julia / master development the last months, so no idea what causes this. Are both tuples no longer inferred? Could it help to add some parametric type declaration to the arguments of the function, to force the compiler to specialise? Maybe there is a heuristic in place by which this function is no longer specialised on the type of its arguments?

dkarrasch:
I'm investigating now, but that is work in progress. It seems like it's puzzling the compiler to do `Base.setindex` on a tuple and rebind the variable to the new tuple. I'm trying to rewrite the `rowcolranges` function using `map`, `ntuple` and broadcasting to construct the needed objects directly.

dkarrasch:
Hm, my solution requires `cumsum` on tuples, which was implemented only in Julia v1.5. How do we feel about lifting the Julia lower bound to v1.6, the assumed next LTS version? After all, this PR fixes an inference issue that came up only on v1.8. Alternatively, we could include `Compat.jl` as a dependency, but that feels strange given that Julia v1.0 has been unmaintained for quite a while. If people agree, I'd use this PR for a little version cleanup.

src/blockmap.jl
```diff
@@ -24,8 +24,8 @@ BlockMap{T}(maps::As, rows::Rs) where {T, As<:LinearMapTuple, Rs} =
 
 MulStyle(A::BlockMap) = MulStyle(A.maps...)
 
-function _getranges(maps, dim, inds::NTuple{N,Int}=ntuple(identity, Val(length(maps)))) where {N}
-    sizes = ntuple(i -> (@inbounds size(maps[inds[i]], dim)), Val(N))
+function _getranges(maps, dim, inds=ntuple(identity, Val(length(maps))))
+    sizes = ntuple(i -> (@inbounds size(maps[inds[i]], dim))::Int, Val(length(inds)))
```
Jutho, line 28: why not `sizes = map(i->size(maps[i], dim), inds)`?

dkarrasch: I'm not sure I have tried this one, but I'm testing many different details, which behave differently in view of type inference in the wide range of Julia versions.

PR #164, opened 2021-11-04
Specialize `Base.axes` for [Composite/Adjoint/Transpose]Maps

dkarrasch:


dkarrasch:
@krcools Do you see any potential issue with this?

krcools:
Thank you for including this also for CompositeMaps. I don't see any issue, apart from maybe we can also get this to work for TransposeMaps and AdjointMaps in a single release.

dkarrasch:
That's difficult to test though, because transpose and adjoint of WrappedMaps of AbstractMatrices are passed to the matrix, so we get a WrappedMap{Transpose{BlockMatrix}}, roughly speaking.

PR #165, opened 2021-12-02
Add slicing functionality

### LINKED ISSUE #38: Why no getindex() ?

dkarrasch:
This is an alternative approach to #145, without adding yet another `LinearMap` subtype. As it stands now, there is no switch for turning on/off indexing/slicing. I was hoping one could have a submodule and "hide" it from the user, unless s/he asks for `using LinearMaps.GetIndex`, but that doesn't seem to work as easy as I was hoping for. OTOH, we might as well trust our users that they are cautious about indexing/slicing and not put additional stones in their way.

This may require a few more tests to keep coverage high, some announcement in the docs and some information on how to extend that functionality to own map types.

Closes #145, closes #38.

### DIFF
--- a/docs/src/history.md
+++ b/docs/src/history.md
@@ -24,6 +24,19 @@
   * `LinearMap(A::MapOrVecOrMat, dims::Dims{2}; offset::Dims{2})`, where the keyword
     argument `offset` determines the dimension of a virtual upper-left zero block, to which
     `A` gets (virtually) diagonally appended.
+* An often requested new feature has been added: indexing/slicing any `LinearMap` object via
+  `Base.getindex` overloads. Linear, Cartesian (both with two `Integer` indices and
+  (vectors of) `CartesianIndex{2}`), colon, and logical indexing as well as slicing all
+  work generically on any `LinearMap` subtype. For performance reasons, some slicing
+  operations require the adjoint operation of the `LinearMap` type to be defined, or will
+  fail otherwise. Important note: `LinearMap` objects are meant to model objects that act
+  on vectors efficiently, and are in general *not* backed up by storage-like types like
+  `Array`s. Therefore, indexing and/or slicing of `LinearMap`s is likely to be slow,
+  and it may require the (repeated) allocation of canonical unit vectors. As a consequence,
+  generic algorithms relying heavily on indexing and/or slicing are likely to work much
+  worse than expected for `AbstractArray`s. To avoid repeated indexing operations which
+  may involve redundant computations, it is strongly recommended to consider `convert`ing
+  `LinearMap`-typed objects to `Matrix` or `SparseMatrixCSC` first.
 
 ## What's new in v3.6
 
--- a/docs/src/history.md
+++ b/docs/src/history.md
@@ -1,6 +1,7 @@
 # Version history
 
 ## What's new in v3.7
+
 * `mul!(M::AbstractMatrix, A::LinearMap, s::Number, a, b)` methods are provided, mimicking
   similar methods in `Base.LinearAlgebra`. This version allows for the memory efficient
   implementation of in-place addition and conversion of a `LinearMap` to `Matrix`.
@@ -9,7 +10,6 @@
   conversion, construction, and inplace addition will benefit from this supplied efficient
   implementation. If no specialisation is supplied, a generic fallback is used that is based
   on feeding the canonical basis of unit vectors to the linear map.
-
 * A new map type called `EmbeddedMap` is introduced. It is a wrapper of a "small" `LinearMap`
   (or a suitably converted `AbstractVecOrMat`) embedded into a "larger" zero map. Hence,
   the "small" map acts only on a subset of the coordinates and maps to another subset of
@@ -24,19 +24,30 @@
   * `LinearMap(A::MapOrVecOrMat, dims::Dims{2}; offset::Dims{2})`, where the keyword
     argument `offset` determines the dimension of a virtual upper-left zero block, to which
     `A` gets (virtually) diagonally appended.
-* An often requested new feature has been added: indexing/slicing any `LinearMap` object via
-  `Base.getindex` overloads. Linear, Cartesian (both with two `Integer` indices and
-  (vectors of) `CartesianIndex{2}`), colon, and logical indexing as well as slicing all
-  work generically on any `LinearMap` subtype. For performance reasons, some slicing
-  operations require the adjoint operation of the `LinearMap` type to be defined, or will
-  fail otherwise. Important note: `LinearMap` objects are meant to model objects that act
-  on vectors efficiently, and are in general *not* backed up by storage-like types like
-  `Array`s. Therefore, indexing and/or slicing of `LinearMap`s is likely to be slow,
-  and it may require the (repeated) allocation of canonical unit vectors. As a consequence,
-  generic algorithms relying heavily on indexing and/or slicing are likely to run much
-  slower than expected for `AbstractArray`s. To avoid repeated indexing operations which
-  may involve redundant computations, it is strongly recommended to consider `convert`ing
-  `LinearMap`-typed objects to `Matrix` or `SparseMatrixCSC` first, if memory permits.
+* An often requested new feature has been added: slicing (i.e., non-scalar indexing) any
+  `LinearMap` object via `Base.getindex` overloads. Note, however, that only rather
+  efficient slicing operations are implemented: `A[:,j]`, `A[:,J]`, `A[I,J]`, and `A[:,:]`,
+  where `j::Integer` and `J` is either of type `AbstractVector{<:Integer>}` or an
+  `AbstractVector{Bool}` of appropriate length ("logical slicing").
+
+  Scalar indexing `A[i::Integer,j::Integer]` as well as other indexing operations that fall
+  back on scalar indexing such as logical indexing by some `AbstractMatrix{Bool}`, or
+  indexing by vectors of (linear or Cartesian) indices are not supported; as an exception,
+  `getindex` calls on wrapped `AbstractVecOrMat`s is forwarded to corresponding `getindex`
+  methods from `Base` and therefore allow any type of usual indexing/slicing.
+  If scalar indexing is really required, consider using `A[:,j][i]` which is as efficient
+  as a reasonable generic implementation for `LinearMap`s can be.
+
+  Furthermore, (predominantly) horizontal slicing operations require the adjoint operation
+  of the `LinearMap` type to be defined, or will fail otherwise. Important note:
+  `LinearMap` objects are meant to model objects that act on vectors efficiently, and are
+  in general *not* backed up by storage-like types like `Array`s. Therefore, slicing of
+  `LinearMap`s is potentially slow, and it may require the (repeated) allocation of
+  standard unit vectors. As a consequence, generic algorithms relying heavily on indexing
+  and/or slicing are likely to run much slower than expected for `AbstractArray`s. To avoid
+  repeated indexing operations which may involve redundant computations, it is strongly
+  recommended to consider `convert`ing `LinearMap`-typed objects to `Matrix` or
+  `SparseMatrixCSC` first, if memory permits.
 
 ## What's new in v3.6
 
--- a/docs/src/history.md
+++ b/docs/src/history.md
@@ -26,9 +26,10 @@
     `A` gets (virtually) diagonally appended.
 * An often requested new feature has been added: slicing (i.e., non-scalar indexing) any
   `LinearMap` object via `Base.getindex` overloads. Note, however, that only rather
-  efficient slicing operations are implemented: `A[:,j]`, `A[:,J]`, `A[I,J]`, and `A[:,:]`,
+  efficient complete slicing operations are implemented: `A[:,j]`, `A[:,J]`, and `A[:,:]`,
   where `j::Integer` and `J` is either of type `AbstractVector{<:Integer>}` or an
-  `AbstractVector{Bool}` of appropriate length ("logical slicing").
+  `AbstractVector{Bool}` of appropriate length ("logical slicing"). Partial slicing
+  operations such as `A[I,j]` and `A[I,J]` where `I` is as `J` above are disallowed.
 
   Scalar indexing `A[i::Integer,j::Integer]` as well as other indexing operations that fall
   back on scalar indexing such as logical indexing by some `AbstractMatrix{Bool}`, or
--- a/docs/src/history.md
+++ b/docs/src/history.md
@@ -33,10 +33,10 @@
   on vectors efficiently, and are in general *not* backed up by storage-like types like
   `Array`s. Therefore, indexing and/or slicing of `LinearMap`s is likely to be slow,
   and it may require the (repeated) allocation of canonical unit vectors. As a consequence,
-  generic algorithms relying heavily on indexing and/or slicing are likely to work much
-  worse than expected for `AbstractArray`s. To avoid repeated indexing operations which
+  generic algorithms relying heavily on indexing and/or slicing are likely to run much
+  slower than expected for `AbstractArray`s. To avoid repeated indexing operations which
   may involve redundant computations, it is strongly recommended to consider `convert`ing
-  `LinearMap`-typed objects to `Matrix` or `SparseMatrixCSC` first.
+  `LinearMap`-typed objects to `Matrix` or `SparseMatrixCSC` first, if memory permits.
 
 ## What's new in v3.6
 
--- a/src/getindex.jl
+++ b/src/getindex.jl
@@ -1,0 +1,132 @@
+# module GetIndex
+
+# using ..LinearMaps: LinearMap, AdjointMap, TransposeMap, FillMap, LinearCombination,
+#     ScaledMap, UniformScalingMap, WrappedMap
+
+# required in Base.to_indices for [:]-indexing
+Base.eachindex(::IndexLinear, A::LinearMap) = (Base.@_inline_meta; Base.OneTo(length(A)))
+Base.lastindex(A::LinearMap) = (Base.@_inline_meta; last(eachindex(IndexLinear(), A)))
+Base.firstindex(A::LinearMap) = (Base.@_inline_meta; first(eachindex(IndexLinear(), A)))
+
+function Base.checkbounds(A::LinearMap, i, j)
+    Base.@_inline_meta
+    Base.checkbounds_indices(Bool, axes(A), (i, j)) || throw(BoundsError(A, (i, j)))
+    nothing
+end
+# Linear indexing is explicitly allowed when there is only one (non-cartesian) index
+function Base.checkbounds(A::LinearMap, i)
+    Base.@_inline_meta
+    Base.checkindex(Bool, Base.OneTo(length(A)), i) || throw(BoundsError(A, i))
+    nothing
+end
+
+# main entry point
+Base.@propagate_inbounds function Base.getindex(A::LinearMap, I...)
+    # TODO: introduce some sort of switch?
+    Base.@_inline_meta
+    @boundscheck checkbounds(A, I...)
+    @inbounds _getindex(A, Base.to_indices(A, I)...)
+end
+# quick pass forward
+Base.@propagate_inbounds Base.getindex(A::ScaledMap, I...) = A.λ .* getindex(A.lmap, I...)
+Base.@propagate_inbounds Base.getindex(A::WrappedMap, I...) = A.lmap[I...]
+Base.@propagate_inbounds Base.getindex(A::WrappedMap, i::Integer) = A.lmap[i]
+Base.@propagate_inbounds Base.getindex(A::WrappedMap, i::Integer, j::Integer) = A.lmap[i,j]
+
+########################
+# linear indexing
+########################
+Base.@propagate_inbounds function _getindex(A::LinearMap, i::Integer)
+    Base.@_inline_meta
+    i1, i2 = Base._ind2sub(axes(A), i)
+    return _getindex(A, i1, i2)
+end
+Base.@propagate_inbounds _getindex(A::LinearMap, I::AbstractVector{<:Integer}) =
+    [_getindex(A, i) for i in I]
+_getindex(A::LinearMap, ::Base.Slice) = vec(Matrix(A))
+
+########################
+# Cartesian indexing
+########################
+Base.@propagate_inbounds _getindex(A::LinearMap, i::Integer, j::Integer) =
+    _getindex(A, Base.Slice(axes(A)[1]), j)[i]
+Base.@propagate_inbounds function _getindex(A::LinearMap, i::Integer, J::AbstractVector{<:Integer})
+    try
+        return (basevec(A, i)'A)[J]
+    catch
+        x = zeros(eltype(A), size(A, 2))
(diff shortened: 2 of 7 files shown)


dkarrasch:
Anyone interested in indexing and slicing `LinearMap`s? 😛 @oschulz 

@JeffFessler: if I forward indexing `LinearMapsA*` to the wrapped `LinearMap`, `LinearMapsAA.jl`'s tests pass for me locally. Could you please double-check?

If this looks good, we could get this in v3.7 soon, otherwise I'd go and make a release with the other currently merged PRs.

oschulz:
> Anyone interested in indexing and slicing LinearMaps?

Very much so! With this, could we do `LinearMap <: AbstractMatrix` (#180) as well, to help with dispatch for "non-LinearMap-aware" code?

dkarrasch:
I'll comment on that in the other issue. For the time being, I'd be interested in getting feedback on the implementation. ;-)

oschulz:
Ok ;-)

src/getindex.jl
```diff
@@ -0,0 +1,143 @@
...
+# specialized methods
+_getindex(A::FillMap, ::Integer, ::Integer) = A.λ
+_getindex(A::LinearCombination, i::Integer, j::Integer) =
+    sum(a -> (@inbounds A.maps[a][i, j]), eachindex(A.maps))
+_getindex(A::AdjointMap, i::Integer, j::Integer) = @inbounds adjoint(A.lmap[j, i])
+_getindex(A::TransposeMap, i::Integer, j::Integer) = @inbounds transpose(A.lmap[j, i])
+_getindex(A::UniformScalingMap, i::Integer, j::Integer) = ifelse(i == j, A.λ, zero(eltype(A)))
+
+# helpers
+function basevec(A, dim, i::Integer)
```
JeffFessler, line 94: Mathematically this is a "standard unit vector" right?   So `unitvec` would be a logical name?
Never mind, this is bike shedding. :)

src/getindex.jl
```diff
@@ -0,0 +1,143 @@
...
+
+########################
+# linear indexing
+########################
+function _getindex(A::LinearMap, i::Integer)
+    i1, i2 = Base._ind2sub(axes(A), i)
+    return _getindex(A, i1, i2)
+end
+_getindex(A::LinearMap, I::Indexer) = [_getindex(A, i) for i in I]
+_getindex(A::LinearMap, ::Base.Slice) = vec(Matrix(A))
```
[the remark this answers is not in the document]
dkarrasch: This has been a while back, and I almost forgot any details here, but the docstring is instructive:

```julia
Slice(indices)

  Represent an AbstractUnitRange of indices as a vector of the indices themselves, with special handling to signal they represent a complete slice of a dimension (:).

  Upon calling to_indices, Colons are converted to Slice objects to represent the indices over which the Colon spans. Slice objects are themselves unit ranges with the same indices
  as those they wrap. This means that indexing into Slice objects with an integer always returns that exact integer, and they iterate over all the wrapped indices, even supporting
  offset indices.
```

Since `_getindex` is "later in the pipeline", at this point we already handle the output of `to_indices`, which is of `Slice` type:

```julia
julia> Base.to_indices(LinearMap(ones(3,3)), (:,))
(Base.Slice(Base.OneTo(9)),)
```

So, the use of `Slice` is not meant to be in support of offset axes, it's really just so we can have a single

```julia
function Base.getindex(A::LinearMap, I...)
    # TODO: introduce some sort of switch?
    @boundscheck checkbounds(A, I...)
    _getindex(A, Base.to_indices(A, I)...)
end
```

and then deal with the output of `to_indices`.

src/getindex.jl
dkarrasch: True, I'll go with that.

dkarrasch:
> When you say "forward indexing" do you mean something like this?
> `Base.getindex(A::LinearMapAX, args...) = Base.getindex(A._lmap, args...)`

Exactly.

oschulz:
> OTOH, we might as well trust our users that they are cautious about indexing/slicing and not put additional stones in their way.

We could use the approach of the GPU packages - warn/error on "serial" access. So we could allow everything that's efficient, like columns and rows (for transposable maps), and warn/error otherwise.

dkarrasch:
> We could use the approach of the GPU packages

@Jutho suggested something very similar, using a macro or something to disallow indexing, or something like a switch. I tried to look up where this machinery (`allowscalarindexing` or so?) is defined, but I just couldn't find out how it works. I'd appreciate any concrete hints here.

oschulz:
> where this machinery (allowscalarindexing or so?) is defined,

It's here: https://github.com/JuliaGPU/GPUArrays.jl/blob/fc0d327ecc2fd0b3b73427cf6f491591aa096b75/src/host/indexing.jl#L23

It uses task-local storage. I guess longer-term, this would be a perfect application for JuliaLang/julia#35833. For now, we could use https://github.com/tkf/ContextVariablesX.jl or use task-local storage like GPUArrays.

JeffFessler:
I did my own test and was able to refer all `getindex` to the methods in this branch and it pretty much worked.
You can see exactly what I had to change here, with commentary:
https://github.com/JeffFessler/LinearMapsAA.jl/pull/48/files
Based on this independent validation I am pretty confident that the methods are correct.

I am unsure though if it is efficient for the case `A[:,j]` because I still don't understand `Slice` enough.
My code translates `A[:,j]` to just `A * univec(A,j)` which I think is as efficient as it gets.
It appears to me that this PR translates it to something like `(A * unitvec(A,j))[1:end]`, if I am understanding the `OneTo` aspect of `Slice`, which seems slightly less efficient than having a dispatch on `Colon` directly.
I won't have time for any timing tests for a while though.

dkarrasch:
> My code translates `A[:,j]` to just `A * univec(A,j)` which I think is as efficient as it gets.

Same here, line 52. (Note, `Slice` is at least as good as `Colon`: it knows that it refers to the entire dimension like `Colon`, but also knows the first and the last indices.)

```julia
_getindex(A::LinearMap, ::Base.Slice, j::Integer) = A*unitvec(A, 2, j)
```

So, I'm just using `Slice` for dispatch, I'm not using its wrapped "value" `Base.OneTo`.

> It appears to me that this PR uses `_fillbycols`

This gets called only when we want `A[I, J]` where both `I` and `J` are `Indexer`s, i.e., `AbstractVector{<:Integer}`. I.e., when you want to have a "submatrix", AND when `length(J)` is smaller than `length(I)`. The latter is to pick the direction of least applications of `A` or `A'`, respectively. This path will get called when you do `A[1:end, 5]`, because `1:size(A,1)` is type-different from `Colon`. Specifically, when you want to get one column, then `J` is not going to be an `Indexer` and the `_fillbycols` is not even in reach.

In your draft, there is one line that says "slow". So this is referring to my implementation? Then I should perhaps find the line in your tests that calls this and benchmark.

JeffFessler:
> there is one line that says "slow"

So that code covers the case of `A[i,:]` for a FunctionMap that has no adjoint function and had I included it for completeness  in my original version.  But in hindsight I think that was unwise for me to include that because that case is *inherently* very slow and it is probably better for that case to fail, to force the user to think about what they are doing.  So I added that comment about it being slow to remind myself of that.  To use this PR as the basis for `getindex` I had to comment out the testset associated with that corner case and I am perfectly fine with deprecating support for that case.

Thanks for the explanation about the dispatch that I hadn't quite followed before.   With that understanding, this PR looks great to me.

docs/src/history.md
```diff
@@ -24,6 +24,19 @@
...
+  `Base.getindex` overloads. Linear, Cartesian (both with two `Integer` indices and
+  (vectors of) `CartesianIndex{2}`), colon, and logical indexing as well as slicing all
+  work generically on any `LinearMap` subtype. For performance reasons, some slicing
+  operations require the adjoint operation of the `LinearMap` type to be defined, or will
+  fail otherwise. Important note: `LinearMap` objects are meant to model objects that act
+  on vectors efficiently, and are in general *not* backed up by storage-like types like
+  `Array`s. Therefore, indexing and/or slicing of `LinearMap`s is likely to be slow,
+  and it may require the (repeated) allocation of canonical unit vectors. As a consequence,
+  generic algorithms relying heavily on indexing and/or slicing are likely to work much
+  worse than expected for `AbstractArray`s. To avoid repeated indexing operations which
```
JeffFessler, line 37:
```suggestion
  generic algorithms relying heavily on indexing and/or slicing are likely to run much
  slower than expected for `AbstractArray`s. To avoid repeated indexing operations which
```

JeffFessler: They will "work" properly...

docs/src/history.md
```diff
@@ -24,6 +24,19 @@
...
+  may involve redundant computations, it is strongly recommended to consider `convert`ing
+  `LinearMap`-typed objects to `Matrix` or `SparseMatrixCSC` first.
```
JeffFessler, line 39:
```suggestion
  `LinearMap`-typed objects to `Matrix` or `SparseMatrixCSC` first, if memory permits.
```

src/getindex.jl
```diff
@@ -0,0 +1,143 @@
...
+_getindex(A::LinearMap, I::Vector{CartesianIndex{2}}) = [(@inbounds A[i]) for i in I]
+_getindex(A::LinearMap, I::Base.LogicalIndex) = [(@inbounds A[i]) for i in I]
+
+########################
+# Cartesian indexing
+########################
+_getindex(A::LinearMap, i::Union{Integer,Indexer}, j::Integer) = (@inbounds (A*unitvec(A, 2, j))[i])
+_getindex(A::LinearMap, ::Base.Slice, j::Integer) = A*unitvec(A, 2, j)
```
JeffFessler, line 52: oh, there it is.  somehow i overlooked this before...

src/getindex.jl
```diff
@@ -0,0 +1,143 @@
...
+function _getindex(A::LinearMap, i::Integer, J::Indexer)
+    # try
+        # requires adjoint action to be defined
```
JeffFessler, line 55: would be helpful here to have a `throw` advising the user to implement an adjoint action instead of just failing on the e'A multiply?  (perhaps there is already such a message in the multiply routine, but even so the user might be thinking they are slicing or indexing not multiplying so a throw message here could still be helpful?

oschulz:
> A[i,:] for a FunctionMap that has no adjoint function and had I included it for completeness in my original version. But in hindsight I think that was unwise for me to include that because that case is inherently very slow and it is probably better for that case to fail

Speaking from a user point of view, I think it would be perfectly fine for all inefficient getindex operations to throw an exception. So no adjoint resulting in no `A[i,:]` is good I think. After all, "inefficient" here typically means "horribly inefficient" - better just not to allow such code to run at all.

src/getindex.jl
dkarrasch: I'm fine with either way. Currently the usual "adjoint not defined for $(typeof(A))" error exception is thrown out of a multiply routine, as you say. Throwing an error out of the getindex routine is perhaps indeed more informative, so users can think about whether they want to "work out" the adjoint, or whether they want to avoid the slicing.

dkarrasch:
> > where this machinery (allowscalarindexing or so?) is defined,
> 
> It's here: https://github.com/JuliaGPU/GPUArrays.jl/blob/fc0d327ecc2fd0b3b73427cf6f491591aa096b75/src/host/indexing.jl#L23
> 
> It uses task-local storage. I guess longer-term, this would be a perfect application for [JuliaLang/julia#35833](https://github.com/JuliaLang/julia/pull/35833). For now, we could use https://github.com/tkf/ContextVariablesX.jl or use task-local storage like GPUArrays.

I think I roughly found that earlier, but had no idea how to translate that to our case here. In other words, what do I need to do to have indexing and slicing disallowed per default, and to force users to turn that on consciously. Also, we would need docs that explain how to do that. I haven't worked with GPUArrays etc. at all, so any concrete contribution would be highly appreciated!

oschulz:
> have indexing and slicing disallowed per default, and to force users to turn that on consciously.

The question is whether we actually want to allow it at all. WIth the GPU arrays the default was changed from serial-allowed to warn-on-serial. I have a feeling that one day it'll be error-on-serial by default. With the warning, you don't directly see which part of code is at fault, so I typically activate `GPUArrays.allowscalar(false)` so I can get an exception with a stack-trace.

As a user I'd don't see a scenario where I'd want to enable inefficient indexing in any real use case.

dkarrasch:
Now I'm confused. What kind of indexing/slicing would you want to have allowed and which ones throwing? With the current state of the PR, if you have an adjoint defined, any indexing/slicing operation is defined (scalar/vector of indices/colon/logical).

oschulz:
> What kind of indexing/slicing would you want to have allowed and which ones throwing?

Personally, I would prefer that anything throws that's "inefficient by construction", so to speak. My suggestion would be any single-element/scalar `getindex`, as well as row-access in the no-adjoint case. Things like `A[2:5,3]` and `A[3,2:5]` are hard to judge, of course, as the efficiency depends on the total number of rows/columns. I would say let's allow that.

The alternative, I think, would be to implement the GPUArray-like warning/error-route, I would suggest the warnings/errors should "trigger" in the same cases.

JeffFessler:
> Personally, I would prefer that anything throws that's "inefficient by construction", so to speak. My suggestion would be any single-element/scalar `getindex`, as well as row-access in the no-adjoint case.

I think I agree, as long as "scalar" means `A[i,j]` or `A[k]` - those are the cases that need warnings (if any).
I consider `A[:,j]` and probably even `A[:,J]` to be fine (i.e., no warning needed).  Let's also keep in mind that `Matrix(A)` is same as `A[:,:]` and there has never been any warning for that case and I'd prefer to retain that behavior.
The general case of `A[I,J]` also reverts to scalar so also could warn.

oschulz:
> I think I agree, as long as "scalar" means A[i,j] or A[k]

Yes, that's what I meant.

> I consider A[:,j] and probably even A[:,J] to be fine (i.e., no warning needed).

I fully agree.

> that Matrix(A) is same as A[:,:] and there has never been any warning

Absolutely - that's an efficient operation, after all.

My inuition would be (with scalar indices `i`, `j` and array/range-indices `I`,`J`) to

* Allow `A[:,j]`, `A[:,J]`, `A[I,J]` and `A[:,:]` in general.
* Only allow `A[i,:]` for maps that support adjoint (need to be careful with complex numbers)
* Disallow `A[i,j]` in general.

`A[I,J]` I would allow in general, I think, to support use cases like `A[begin+1:end-1:begin+1:end-1]` and similar, which would typically still very efficient.

oschulz:
> The latest commit largely implements the suggestions (thanks for the clear roadmap!). 

Thanks a lot for doing all of this!

> What I currently have errors on A[i,j], and suggests to consider using A[:,j][i] if one really needs to.

Oh, that's a nice approach I think!

> Question: shall we do the analogous thing for A[I,j], i.e., error and suggest A[:,j][I]? ...  I wonder what rationale exactly applies to allowing A[I,J] but not A[i,j]

My intuition was that code which iterates over `A[I,j]` with short ranges of `I` is comparatively rare, while stuff like `A[begin+1:end-1,j]` is not uncommon (getting the inner part things that were padded to fold with kernels and so on). The same for `A[I,J]`.

> `A[begin+1:end-1,begin+1:end-1]` may not be more efficient than `convert(Matrix, A)[begin+1:end-1,begin+1:end-1]`

Yes, I think so too. I was assuming that the user code wouldn't want to have the actual matrix in hand, necessarily. Maybe it would in fact be best if `A[I,J]` would not evaluate anything be return something like a view? But the implementation of `A[begin+1:end-1,begin+1:end-1] * x` wouldn't be trivial - one would have to pad `x` first, to the mul, and then do a getindex on the result.

Maybe it's best if we disallow `A[I,j]`, `A[i,J]` and `A[I,J]` for now (error with suggestion of alternative code, you you already have) until we come across valid use cases?

dkarrasch:
> Maybe it's best if we disallow `A[I,j]`, `A[i,J]` and `A[I,J]` for now (error with suggestion of alternative code, you you already have) until we come across valid use cases?

Yes, I guess that's much better than first allow it and then handle potential "Why is XYZ so slow?" issues... or even disallow it later. I think the main point that the user needs to understand is that these cases do correspond to first compute entire slices and then slice again. If users will have to write that out they will naturally want to think twice if that's really what they want, or whether they we're writing it just out of convenience.

JeffFessler:
Makes sense. 
Then probably `A[:,:]` also should throw for now, whereas `Matrix(A)` continues to work silently (and possibly slowly)?

dkarrasch:
Ha, that will make a fantastic PR: basically add a whole bunch of methods that throw! Progress! Yippieh! 🤣

oschulz:
> Then probably A[:,:] also should throw for now,

I think `A[:,:] ` should be allowed. It's very efficient, and it would be surprising for the user if `A[:,i]` is allowed but `A[:,:]` isn't.

JeffFessler:
> It's very efficient

I guess that is in the eye of the beholder.  If `A` is the N-point DFT implemented via an FFT (a typical use case for me), then `A*x` is $O(N \log N)$ whereas `A[:,:]` is $O(N^2 \log N)$ which is more expensive than the $O(N^2)$ needed to simply construct the matrix in the first place!  So my view is that it should be discouraged a bit (if we are also discouraging `A[I,J]`), but I don't feel too strongly about it.
Perhaps ideally there would be a user switch (like with GPU scalars) but it's not a high enough priority for me to want to try to figure that out.

oschulz:
> then A*x is P(N log N) whereas A[:,:] is O(N^2 log N)

Yes, that's what I mean - it scales proportionally to the number of outputs that you want to produce, both for `A[:,i]` and `A[:,:]`. An overhead of log(N) is typically still considered quite efficient in numerics, after all.

I would put it like this: `A[:,i]` is one mul for an output vector with N numbers - that's definitely Ok, it's the bare performance of the linear map. And `A[:,:]` is N muls for an output matrix of N^2 numbers. That's very Ok too, I would say.

dkarrasch:
For higher-dimensional arrays, I guess it is not completely uncommon to put colons in two dimensions. For "matrices", who would construct a (deep) copy via `A[:,:]`, when you have `Matrix(A)`?

BTW, `Matrix(A)` was never meant to be "efficient". It was just introduced so that, if you really need (for instance, for stubborn upstream packages 😝 ) to provide a matrix. One often repeated argument by @Jutho is that, for applications that require indexing of many different components, it may be beneficial to pay the full (matrix construction) price once than to have it potentially an unbound number of allocations etc. For such cases, we have `Matrix(A)` and `sparse(A)` constructors. So, not "efficient", but perhaps "better than". For that reason, `Matrix(A)` was never viewed in terms of indexing/slicing. Now, with this PR, these two things (matrix construction and slicing) touch each other via `A[:,:]`, and one may take two perspectives: (a) construction of a matrix, and (b) double-slicing. From the perspective (a), "efficiency" doesn't play a role, it's just supposed to be "better than"; from the perspective (b), this is an awful operation, and should be warned about, to the very least. Obviously, it's just a matter of taste, but I tend to prefer path (b). Note that generic code could use `convert(AbstractMatrix, A)`, which is a noop for `AbstractMatrix`es.

oschulz:
> For "matrices", who would construct a (deep) copy via A[:,:], when you have Matrix(A)

It's not quite the same in general, because for normal arrays ` A[:,:]` would typically preserve the array type. For LinearMaps it won't matter much which is used, I guess.

My argument for allowing `A[:,:]` is more that we can easily do it and that allowing `A[:,i]` (which we want) makes forbidding `A[:,:]` (which has the same relative efficiency) surprising for the user.

dkarrasch:
Ok, the rationale of allowing "any complete slicing" (as opposed to partial slicing, which requires more allocations than just the unavoidable unit vector and the result) sounds convincing to me.

As I said, it's borderline and a matter of taste: from the point of view of matrix construction it's clear that this is expensive (and could deserve a warning), from the point of view of slicing it is "expectedly expensive" (and could therefore be allowed). I'll see if I need to adjust the announcement. Perhaps, there should also be a section in the `types.md` docs, that repeats considerations and lays out the rationale of what is allowed and what not.

dkarrasch:
Uff, I think I finally understood the macro stuff! 🎉 

What shall we do then? Allow slicing in general without interference of macros and allow scalar indexing/partial slicing upon a macro signal, or allow any slicing/indexing only upon macro signal? Having two macros, one for slicing and one for indexing and partial slicing is perhaps overkill.

oschulz:
> What shall we do then? Allow slicing in general without interference of macros

From a user perspective, I'd be happy with `A[:,j]` and `A[:,:]` allowed, `A[i,:]` allowed if adjoint exits, and everything else resulting in an error with a helpful "don't do this, it's slow" error message. Anything beyond (macro magic to allow stuff explicitly, etc.) is fine too, of course.

dkarrasch:
Ok, then you should be happy with the current state, I hope. We could release this as is and then take some time to work out how to allow things that are currently forbidden. That would be "a new feature" and non-breaking, as opposed to first allowing something and later require a macro to do that. I'll wait for @JeffFessler's thumb.

oschulz:
Sounds great - again, thanks a lot!

src/getindex.jl
```diff
@@ -0,0 +1,113 @@
+const Indexer = AbstractVector{<:Integer}
+
+Base.IndexStyle(::LinearMap) = IndexCartesian()
+# required in Base.to_indices for [:]-indexing
+Base.eachindex(::IndexLinear, A::LinearMap) = Base.OneTo(length(A))
+Base.lastindex(A::LinearMap) = last(eachindex(IndexLinear(), A))
+Base.firstindex(A::LinearMap) = first(eachindex(IndexLinear(), A))
```
JeffFessler, line 7: I noticed that codecov complains about these two lines.  I wonder if they should be commented out because we are not supporting `A[1]` and `A[end]` at this point in time, right?

src/getindex.jl
```diff
@@ -0,0 +1,113 @@
...
+end
+
+# helpers
+function unitvec(A, dim, i)
+    x = zeros(eltype(A), size(A, dim))
+    @inbounds x[i] = one(eltype(A))
+    return x
+end
+
+function _fillbyrows!(dest, A, I, J)
```
JeffFessler, line 83: should these be commented out since `A[I,J]` is not supported?
the codecov warning is what caught my eye here.

dkarrasch: The tests were not variable enough. This method is needed in general.

JeffFessler:
> I'd be happy with `A[:,j]` and `A[:,:]` allowed, `A[i,:]` allowed if adjoint exits

this is also fine with me.  thumbs up other than a couple questions i asked about commenting out possibly unused code which i leave to you to decide.

oschulz:
Thanks again!

PR #167, opened 2022-01-24
Fix links to documentation

tknopp:
they were pointing to nowhere

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,11 +30,11 @@ julia> import Pkg; Pkg.add("LinearMaps")
 [docs-stable-img]: https://img.shields.io/badge/docs-stable-blue.svg
 [docs-stable-url]: https://JuliaLinearAlgebra.github.io/LinearMaps.jl/stable
 
-[build-img]: https://github.com/Jutho/LinearMaps.jl/workflows/CI/badge.svg?branch=master
-[build-url]: https://github.com/Jutho/LinearMaps.jl/actions?query=workflow%3ACI+branch%3Amaster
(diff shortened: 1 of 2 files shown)


dkarrasch:
Thanks. That got messed up when moving the package to JuliaLinearAlgebra. I assume now any reference to @Jutho's account is wrong, isn't it? I see more of those references in the CI/build status and code coverage in the README and in the doc's `make.jl`. Would you be able to fix them as well?

tknopp:
did not scroll further down :-)

dkarrasch:
... and the `docs/make.jl` file, please. In the `deploydocs` function, there is another `/Jutho/LinearMaps.jl` where `Jutho` needs to be replaced by `JuliaLinearAlgebra`.

dkarrasch:
I guess the documenter key needed an update. Let's see if it worked.

dkarrasch:
Hm, the problem is that the documentation is not deployed because something is wrong with the deploy keys (though I tried to replace the keys with the new "owner" name `JuliaLinearAlgebra`). Does anybody know what adjustments need to be made after moving a repo? Maybe @ViralBShah knows (somebody who knows)?

tknopp:
Do you expect deploying to work on a PR?

dkarrasch:
I believe it used to deploy. We had a preview link to the new documentation in that box listing all the GHA steps. But you may be right. I could merge this and see what happens.

dkarrasch:
Thanks for pushing me! 😝  Seems like it's working.

PR #171, opened 2022-02-28
Allow vectors for LinearMap storage

### LINKED ISSUE #169: Support for arrays in addition to tuples in LinearCombination
We use LinearMaps extensively in [MGVI.jl](https://github.com/bat/MGVI.jl). We're adding support for Newton-CG-optimization now which needs means of linear maps. The number of maps in the mean/sum is not fixed though, so we'd like to use an array of maps instead of a tuple in `LinearCombination`. Would a pull request to extend `LinearCombination` that way be welcome?

dkarrasch:
This extends the `maps` fied of higher-order `LinearMap`s to be a vector, rather than only a tuple. It does not change default behavior. To construct such maps, one would need to call non-exported constructors directly ("expert use").

Closes #169. @oschulz


src/LinearMaps.jl
```diff
@@ -69,6 +73,21 @@ convert_to_lmaps(A) = (convert_to_lmaps_(A),)
...
+_front(As::Tuple) = Base.front(As)
+_front(As::AbstractVector) = @inbounds @views As[1:end-1]
+_tail(As::Tuple) = Base.tail(As)
+_tail(As::AbstractVector) = @inbounds @views As[2:end]
+
+_combine(A::LinearMap, B::LinearMap) = tuple(A, B)
+_combine(A::LinearMap, Bs::LinearMapTuple) = tuple(A, Bs...)
+_combine(As::LinearMapTuple, B::LinearMap) = tuple(As..., B)
+_combine(As::LinearMapTuple, Bs::LinearMapTuple) = tuple(As..., Bs...)
+_combine(A::LinearMap, Bs::LinearMapVector) = vcat(A, Bs...)
```
Jutho, line 85: I don't think you need to splat vector arguments in `vcat`. This could lead to calling `vcat` with a large number of arguments, and thus trigger additional compilation.

dkarrasch: Yes, of course, that was oversight!

dkarrasch: I originally meant to not splat, so this was oversight. But now I benchmarked, and it turns out that splatting has a very positive effect for both short (5)  and long (500) vectors. For instance:

```julia
julia> @btime vcat(x..., y...) setup=(x = (rand(500)...,); y = rand(500));
  47.389 μs (1003 allocations: 35.50 KiB)

julia> @btime vcat(x..., y) setup=(x = (rand(500)...,); y = rand(500))
  11.418 ms (259012 allocations: 5.98 MiB)

julia> @btime vcat(x..., y...) setup=(x = (rand(5)...,); y = rand(5))
  345.180 ns (12 allocations: 352 bytes)

julia> @btime vcat(x..., y) setup=(x = (rand(5)...,); y = rand(5))
  1.230 μs (37 allocations: 944 bytes)
```

oschulz:
Thanks a lot for this, @dkarrasch! Would you consider adding a specialization

```julia
Base.sum(maps::LinearMapTupleOrVector) = LinearCombination(...)
```

to this PR? It would be very helpful for our current use case in MGVI.jl.

dkarrasch:
That already works via the `+(::LinearMap, ::LinearMap)` fallback. I'll add that as a test case. Somewhere in the tests, we already have a corresponding `prod` test.

src/LinearMaps.jl
dkarrasch: Ha, it turns out that this was the wrong approach anyway. We need `Base.vect` instead of `vcat`, because the latter constructs a `BlockMap`.

oschulz:
> That already works via the +(::LinearMap, ::LinearMap) fallback 

It does, and we're using it, but it results in a tuple-backed `LinearCombination` which results in a high compile-time cost if the number of elements in the sum is high. In our case, it's also variable, so many different `LinearCombination` types get created.

Specializing sum would result in an sum of an array of linear maps being turned into an array-backed `LinearCombination`. That would be very helpful to our use case (and potentially other too, I would hope).

oschulz:
>> the promoted eltype is not inferable.
> I'm not quite sure I got that 

Ah, sorry, it's about `T`. True, in the general case it's not inferable. But I think a very typical use case will be a sum over linear maps with the same element type. We could make it inferable/type-stable for that case at least. And operations on arrays with mixed element types aren't inferable anyway.

dkarrasch:
So, do you think you'll need an overload of `mean` as well? Currently it works on vectors of linear maps, but doesn't respect the vector backend.

oschulz:
> So, do you think you'll need an overload of mean as well?

Having `mean` would be great!

oschulz:
Saw your commits regarding `mean` - thanks a lot!

oschulz:
Again, thanks a lot for this @dkarrasch ! Will this be v3.5.2 or v v3.6.0?

oschulz:
We can definitely run some comparison tests with MGVI.jl.

CC @apmypb, @jknollm, @vindex10

cvsvensson:
> To construct such maps, one would need to call non-exported constructors directly ("expert use").

These maps can also be constructed using Base.prod, which leads to an indexing problem if the number of maps is one.
```julia
using LinearMaps
m = LinearMap(cumsum,3)
typeof(m*m) #Tuple type
typeof(prod([m,m])) #Vector type
typeof(prod([m])) #Vector type, and leads to an indexing problem:
prod([m])*ones(3) #ERROR: BoundsError: attempt to access 1-element Vector{LinearMaps.FunctionMap{Float64, typeof(cumsum), Nothing}} at index [2]
```

The error originates from here:
https://github.com/JuliaLinearAlgebra/LinearMaps.jl/blob/f8eaf9846f865f74ddb5dda831081f6a6e8db734/src/composition.jl#L207-L211

Issue #134, opened 2021-01-27
RestrictionMap/SamplingMap?


dkarrasch:
I was reading carefully the PyLops paper that @goretkin mentioned in #98, and noticed that there might be some interest in operators that restrict/sample vectors, `y = x[I]`, whose transpose/adjoint can be interpreted as an embedding `y[I] .= x`. Interestingly, in that paper they show that their sampling operator is basically as fast as the corresponding multiplication with the corresponding sparse matrix (the authors even compare to the multiplication with a dense sampling matrix and seem to make the case that avoiding to store the matrix is of great benefit). I was curious if we can both avoid storing the matrix _and_ have faster performance compared to the sparse matrix approach, as it should be! Of course, this has nothing to do with `LinearMaps.jl`, but all with array indexing handled by Julia Base, but still, and it's possible, of course.

This is all trivial and I don't want to blow up this package unnecessarily. But if there's interest, this could be a cute addition. 

This is somewhat related to #65, but cannot replace it, since we wouldn't want to represent an `IndexMap` as the product `E*A*R` of a transpose restriction map/embedding map `E`, a linear map `A`, and a restriction map `R` (to avoid the intermediate "copies"), but have the indexing included. Thumbs up or down for adding it?

dkarrasch:
Actually, this would play quite nicely with `IndexMap` (pseudocode):
```julia
struct RestrictionMap <: LinearMap{Bool}
    ...
end
struct EmbeddingMap <: LinearMap{Bool}
    ...
end

transpose(R::RestrictionMap) = EmbeddingMap
transpose(R::EmbeddingMap) = RestrictionMap

Base.:(*)(E::EmbeddingMap, A::LinearMap, R::RestrictionMap) =
    IndexMap(A, (size(E, 1), size(R, 2)), E.indices, R.indices)
Base.:(*)(A::LinearMap, R::RestrictionMap) =
    IndexMap(A, (size(A, 1), size(R, 2)), 1:size(A, 1), R.indices)
Base.:(*)(E::EmbeddingMap, A::LinearMap) =
    IndexMap(A, (size(E, 1), size(A, 2)), E.indices, 1:size(A, 2))
```

dkarrasch:
@JeffFessler Do you think there's value in this in `LinearMaps.jl`? Not sure which other application cases might use it.

JeffFessler:
This kind of sampling is exactly what I use for MRI reconstruction:
[...]

PR #174, opened 2022-04-04
Add `EmbeddedMap`

dkarrasch:
This is a rebase of #65. Closes #65.

Dropping the requirement of monotonically increasing indices would allow to define "lazy" reversed maps, though negative strides in BLAS.gemv! is allowed starting from v1.8 only.

I wonder if `IndexMap` is the best name for it, though. "Index" is fairly generic, so we may want to spend some brain power to come up with a better name.


dkarrasch:
If I'm not mistaken, then this PR coincides greatly with [`LiftedMaps.jl`](https://github.com/krcools/LiftedMaps.jl), so @krcools may want to take a look and make suggestions.

dkarrasch:
Another thing: if one ignores the monotonic indices test, then this allows for lazy `reverse`al of matrices/maps! For matrices, that requires negative strides in BLAS, which is available only from v1.8, but still, another nice lazy feature.

dkarrasch:
@krcools What do you think: I'm finding the name `IndexMap` too generic, so I was searching for other names, ideally with some precedence. Then I found your `LiftedMap`s, that has essentially the same functionality, and I assume some precedence regarding the name. What are your plans with your package? Do you think there is a possibility that we name the maps defined here as `LinearMaps.LiftedMap`s? Or shall we keep things independent and search for another name? It would be good to finish this and release it soon as v3.7, together with some of the other open PRs.

krcools:
I am using LiftedMaps in my own registered projects and will keep doing so.  Functionality is indeed similar; a noteworthy difference is that LiftedMaps tries to be compatible with non-traditional axes (such as the blocked axes from BlockArrays).

I guess if the name is not exported both implementations can happily coexist.

Alternative names? Perhaps `ExtendedMaps`, `EmbeddedMaps`, `SingleBlockMaps`?

dkarrasch:
Thanks for your feedback @krcools! Yes, we have avoided the dependency on BlockArrays.jl when we added the `BlockMap` functionality. I think I'll go with `EmbeddedMap`, that comes close to what I suggested in #134, which was closed in favor of the predecessor of this PR. Any name requires explanation, but `EmbeddedMap` comes with a little bit more context than just `IndexMap`.

PR #177, opened 2022-04-06
AD support via chainrules

### LINKED ISSUE #176: AD support
Is there a plan to support (partial) AD for Flux for example? Adding `ChainRulesCore` rules for `A*x` should be fairly easy and would allow using LinearMaps easily in ML workflows (considering A non-differentiable since it's a matrix-free operator).

Something similar to:

https://github.com/slimgroup/JOLI.jl/pull/32

mloubout:
Supports AD via chainrules for the multiplication, closes #176.

Currently, `mul!` does not support AD because I'm not sure how to define a perturbation on the LinearMap itself for the forward rule.

Added a test as well.


Project.toml
```diff
@@ -3,9 +3,11 @@ uuid = "7a12625a-238d-50fd-b39a-03d52299707e"
 version = "3.6.1"
 
 [deps]
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
```
mloubout, line 6: FYI this package only depends on `Compat`, `SparseArrays` and `LinearAlgebra` so it's quite small.

src/LinearMaps.jl
```diff
@@ -10,6 +10,8 @@ using SparseArrays
 
 import Statistics: mean
 
+import ChainRulesCore: rrule, frule, unthunk, @thunk, NoTangent
```
dkarrasch, line 13: I think you need to `import` only those functions that you wish to overwrite, in this case `rrule`. As for the rest, I think `using ChainRulesCore` should be fine, or, if you want to be more restrictive, then `using ChainRulesCore: unthunk, @thunk, NoTangent`. `frule` doesn't seem to be used anywhere.

src/chainrules.jl
```diff
@@ -0,0 +1,19 @@
...
+
+function rrule(A::LinearMap, x::AbstractVecOrMat)
+    y = A(x)
+    function pullback(dy)
+      DY = unthunk(dy)
+      # Because A is an abstract map, the product is only differentiable w.r.t the input
+      return NoTangent(), @thunk(A' * DY)
+    end
+    return y, pullback
+end
```
dkarrasch, line 19: Please add a new line here. Same below in the test file.

dkarrasch (commented):
Thanks for the clear contribution. I just have a few style remarks. As this is unfamiliar ground for me, is there any chance to have special differentiation rules for `FunctionMap`s that would "expose" the code to `Flux`? I have no idea what people need or use, but I imagine that would be awesome. Our usual toy example is `cumsum`, which in terms of a matrix is just `LowerTriangular(ones(m,m))`. That doesn't have parameters in it so the derivative should be zero. But one could imagine a "weighted cumulative sum" with some weight vector input. Since for any weight we obtain a linear map, taking derivatives shouldn't lead us out of the linear map realms, naively speaking? Or the other way around, forgetting about `LinearMaps.jl` and just considering "weighted cumsum" as a piece of code, that should be accessible to AD, shouldn't it? Anyway, I'm just thinking out loud without any concrete idea.

mloubout:
Thanks for the quick review.

Concerning `FunctionMaps` this is not really a trivial thing. For some special case you can write a specific rule, for example there is rules defined for `Diagonal` in `ChainRules.jl` but that's on a case-by-case basis and cannot really be generalized to an anonymous function. 
So conceptually yeah that would be great, but it's a lot more work than this PR. I

Jutho:
What are the problems with AD not handling this (not sure what "this" is) automatically?

I am currently experimenting with AD in the context of KrylovKit.jl, for the solution of linear problems or eigenvalue problems. In KrylovKit, the linear operator is generically specified as a function (not a `LinearMap`). It seems possible to define the rules even for the case where these are functions with internal parameters, so assuming `A` to be non-differentiable is not even necessary. But there the hard part is the rule for the linear problem/eigenvalue problem itself; the rule for the matrix vector product should be easy enough that AD can handle it by itself, no?

dkarrasch:
Thanks for chiming in, @Jutho. And indeed, the provided tests almost work, they fail with

```julia
Can't differentiate gc_preserve_end expression
  Stacktrace:
    [1] error(s::String)
      @ Base ./error.jl:35
    [2] Pullback
      @ ~/Documents/julia/usr/share/julia/stdlib/v1.9/LinearAlgebra/src/blas.jl:644 [inlined]
```

from within BLAS. But with `A = LinearMap(cumsum, reverse∘cumsum∘reverse, 10)` instead, all tests just pass even without this PR. 🎉  OTOH, I quickly tested with a `BlockMap`, and then it fails again with

```julia
Mutating arrays is not supported -- called copyto!(::SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, _...)
```

So maybe this PR is necessary to cover up internal "issues"?

mloubout:
(sorry for the second message)

Finally, defining the rule internally that `d (Fx)/ dx = F'` (since we know it's linear operator with defined adjoint) may lead to more straightforward AD as for example `everse∘cumsum∘reverse,` will be treated as three different function that may degrade performance (compute and memory)

Project.toml
```diff
@@ -3,9 +3,11 @@ uuid = "7a12625a-238d-50fd-b39a-03d52299707e"
 version = "3.6.1"
```
dkarrasch, line 3:
```suggestion
version = "3.7.0"
```

dkarrasch (commented):
Since this is a new feature, please bump the minor version, and please write a short "news" entry in the documentation (`history.md`), then we're ready to go.

dkarrasch:
For reasons of performance and functionality (handling of inplace operations etc.), this seems like a necessary step. Shall we merge this? @Jutho Were you afraid that this PR may prevent getting derivatives of the linear operator working later? Perhaps, if we have concrete use cases, we may add more specific rules, like directing `dA/dp` where `A` is the linear map and `p` is some set of parameters to the wrapped map?

mloubout:
Sorry for the delay, I am on parental leave will get back to it asap.

IMO there is two main cases
- A standard linear function (i.e cumsum). In that case the function is fixed so it's linear representation is as well so ` dA/dx` doesn't  exist
- A parametric function. In that case it may not be linear anymore depending  on the parameteization so it would probably be on a case by case with a specific rule for a given parametric map.

Happy to iterate over the PR to get it to a best usable point

Jutho:
As I am no longer actively using this package, I don't think my opinion should be valued highly.

That being said, I would indeed like to see use cases where this rule really helps, where there is no custom rule higher up that states that you need to do `A' * x`, and where you are not interested in varying parameters in `A`. For example, the case I am most familiar with where LinearMaps are used is in Krylov methods. However, I think it would be a bad idea to AD through Krylov methods, i.e. you will want a custom `rrule` for the Krylov method itself, which will be calling the action of `A'` . But there are likely many use cases that I am not at all familiar with.

Assuming that just because a function has additional parameters would make it a nonlinear function of `x` is not really a good argument. It would be rather evil to insert a non-linear function into `LinearMap`, and the consequences of that would be the user's own responsibility.

mloubout:
Sorry for the delay will get to it next week.

src/chainrules.jl
```diff
@@ -0,0 +1,19 @@
...
+    function pullback(dy)
+      DY = unthunk(dy)
+      # Because A is an abstract map, the product is only differentiable w.r.t the input
+      return NoTangent(), NoTangent(), @thunk(A' * DY)
+    end
+    return y, pullback
+end
+
+function rrule(A::LinearMap, x::AbstractVecOrMat)
+    y = A(x)
```
JeffFessler, line 12:
```suggestion
    y = A*x
```
Perhaps it will work as is, but I recommend we keep reminding everyone that this is a linear map by using multiplication rather than a function call...

src/chainrules.jl
```diff
@@ -0,0 +1,19 @@
+function rrule(::typeof(*), A::LinearMap, x::AbstractVecOrMat)
```
JeffFessler, line 1: It looks like the tests consider only the case where `x` is a vector.
Because `A * X` returns a new `LinearMap` here when `X` is a matrix, it is a bit unclear to me whether this will work, so I'd recommend either restricting it to vectors or explicitly testing the case where `x` is a matrix.

JeffFessler:
I'd like to add my general voice of support for building in support for autodiff here.
I believe `ChainRulesCore` is indeed the appropriate mechanism, but am not completely sure about that because I've never tried it myself.
My group has a `LinearMapAO` for SPECT that is pretty efficient by itself, but autodiff of `A*x` was very slow until we wrote a custom `@adjoint` using `Flux`
https://github.com/JeffFessler/SPECTrecon.jl/blob/7ae2b750fafdd8baa4a7ad0d5399cbe9f62e31ee/docs/lit/examples/6-dl.jl#L167
It would be convenient to have capability this built-in so that we don't need to repeat that for every new `LinearMap`.

src/chainrules.jl
```diff
@@ -0,0 +1,19 @@
...
+    y = A*x
+    function pullback(dy)
+      DY = unthunk(dy)
+      # Because A is an abstract map, the product is only differentiable w.r.t the input
+      return NoTangent(), NoTangent(), @thunk(A' * DY)
```
JeffFessler, line 6: This will work only for maps that have an adjoint.
To help out users, I recommend a try/catch like we have for `getindex` here
https://github.com/JuliaLinearAlgebra/LinearMaps.jl/blob/99bbc0e46a196dee503bffee4612bc5494656cd3/src/getindex.jl#L54

JeffFessler:
Bump?  I think there will be increasing interest in automating AD for linear maps:
https://github.com/JuliaMath/NFFT.jl/issues/50

tknopp:
Out of curiosity: Is there a reason to use `Flux` in the tests instead of `Zygote`? If I see it right, only the `gradient` function from `Zygote` is used. (There is also  https://juliadiff.org/ChainRulesTestUtils.jl/stable/)

src/chainrules.jl
dkarrasch: I'm not sure either how to squeeze in a try-catch block, but I tried and you get the usual "transpose/adjoint not defined for [map type]" error message.

mloubout:
I am quite late on my TODOs, I'm really sorry this is taking a while.

>Is there a reason to use Flux in the tests instead of Zygote?

Just the habit, no particular reason and I'm fine switching to Zygote.


> This will work only for maps that have an adjoint.

This is the assumption for a Linear Operator yes when used as part of a "backpropagated" code. If the adjoint is not defined then an error should occur but I am not sure what the "cleanest" way to do so is to avoid having it mized in the middle of Zygote error messages

Project.toml
dkarrasch: Yea, the only issue seems to be that it adds quite a number of (potential) method invalidations, but I guess this is something people are working on over there.

mloubout:
>could add differentiation w.r.t. the operator later, right?

Yes. I think it would go the `Tangent{LinearMap}(;.....)` way that would follow the design of ChainRules (and the update rule accordingly). It would require some more careful case-by-case design I think but could maybe be generalized. But as you said, the current state allow to easy integration of "static" operators within ML code which is a step in a good direction IMO.

I'll work on the tiny Doc update hopefully by end of the week max. Thanks for the patience

src/LinearMaps.jl
```diff
@@ -11,6 +11,9 @@ using SparseArrays
 
 import Statistics: mean
 
+using ChainRulesCore
```
JeffFessler, line 14:
```suggestion
using ChainRulesCore: unthunk, NoTangent, @thunk
```
It helps others know where to look for odd words like "thunk".  Up to you...

JeffFessler:
> somebody is already re-implementing LinearMaps.jl
sigh ☹️
Otherwise, LGTM other than one small code suggestion to aid future readers.

docs/src/history.md
```diff
@@ -2,6 +2,11 @@
 
 ## What's new in v3.8
 
+* Linear Maps are now differentiable w.r.t to the input via Chainrules for integration with
+  machine learning framewiorks such as [Flux.jl]. The reverse differentiation rule is definded such that
```
JeffFessler, line 6:
```suggestion
  machine learning frameworks such as [Flux.jl]. The reverse differentiation rule is definde such that
```

mloubout:
Thank you for the patience, I think all comments are answered now.

Issue #185, opened 2022-06-22
Type issue with `kron`


JeffFessler:
It seems that `kron` may not be inheriting the `eltype` of component linear maps:

```julia
using FFTW: fft
using LinearMaps
M,N = 64, 32 # image size
T = ComplexF64
Ax = LinearMap(T, x -> fft(x), M, M)
Ay = LinearMap(T, x -> fft(x), N, N)
Ak = kron(Ay, Ax) # has wrong eltype (non-complex)
Ak * rand(T, M*N) # fails with MethodError, perhaps due to type issue?
```

I was working on a suggestion for #183 when I hit this issue.

dkarrasch:
The `eltype` parameter needs to be in curly brackets, otherwise it will interpret `T` as the "forward" map, and `fft` as the adjoint. 😄 

```julia
using FFTW: fft
using LinearMaps
M,N = 64, 32 # image size
T = ComplexF64
Ax = LinearMap{T}(fft, M, M)
Ay = LinearMap{T}(fft, N, N)
Ak = kron(Ay, Ax) # has wrong eltype (non-complex)
Ak * rand(T, M*N) # fails with MethodError, perhaps due to type issue?
```

JeffFessler:
Oh gosh, I am so used to seeing `T` as the first (optional) type argument in the Julia ecosystem, e.g., `ones(T, ...)` and `rand(T, ...)`, that I forgot the `{T}` syntax for LM.  Doh!

But I am surprised that `A = LinearMap(Float64, cumsum, 8, 8)` even works and returns a `FunctionMap`.  I'm proposing #186 so that next time anyone makes this mistake they will get an error at construct time, instead of a useless `FunctionMap`.

I realize that a type actually is a kind of function, cf, `Float32(3)`, even though `Float32 isa Function` is `false`.  Hard to imagine a use case for a type as the function in a FunctionMap though...

PR #186, opened 2022-06-22
Require `FunctionMap` to be passed functions

JeffFessler:
Untested...
Hopefully would address secondary issue buried within #185.

### DIFF
--- a/src/functionmap.jl
+++ b/src/functionmap.jl
@@ -12,7 +12,7 @@ function FunctionMap{T}(f::F1, fc::F2, M::Int, N::Int;
     ismutating::Bool  = _ismutating(f),
     issymmetric::Bool = false,
     ishermitian::Bool = (T<:Real && issymmetric),
-    isposdef::Bool    = false) where {T, F1, F2}
+    isposdef::Bool    = false) where {T, F1 <: Function, F2 <: Function}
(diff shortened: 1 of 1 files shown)


fredrikekre:
This would disallow callable types, right? You would need to pass `LinearMap(x -> A(x))` after this change.

JeffFessler:
> This would disallow callable types, right? You would need to pass `LinearMap(x -> A(x))` after this change.

Right.  Are there use-cases for that?  (The CI doesn't test for them.)
It's OK to close this little PR if the preference is to stay with duck typing.

dkarrasch:
I'm not sure we want this. First, it's technically breaking, and by callable types, one can image, for instance, `LinearMap`s. They get special treatment and get wrapped, but other people could have

```julia
struct CumSum
    n::Int
end
CumSum(x::AbstractVector) = cumsum(x) # up to size checking
```

to adopt our favourite example from the tests. In fact, I should perhaps add such a callable type to the tests.

Second, the `eltype` parameter in curly brackets has important precedence in the Julia language, and its usage is very similar to ours:

```julia
x = Vector{Int}(undef, 10)
```

The reason is that `T` is the first type parameter, and those are always written in curly brackets after the constructor.

JeffFessler:
ok

PR #190, opened 2022-09-01
Add Invalidations.yml

ranocha:
This is based on https://github.com/julia-actions/julia-invalidations. Adding such checks came up in https://discourse.julialang.org/t/potential-performance-regressions-in-julia-1-8-for-special-un-precompiled-type-dispatches-and-how-to-fix-them/86359. I suggest to add this check here since this package is widely used as a dependency.

See also SciML/MuladdMacro.jl#26 and SciML/MuladdMacro.jl#29

### DIFF
--- a/.github/workflows/Invalidations.yml
+++ b/.github/workflows/Invalidations.yml
@@ -1,0 +1,40 @@
+name: Invalidations
+
+on:
+  pull_request:
+
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: always.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  evaluate:
+    # Only run on PRs to the default branch.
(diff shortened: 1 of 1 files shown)


dkarrasch:
Wow, thanks, what a service. Unfortunately, CI doesn't run on this. Besides the mentioned PRs above, do you have a case where one can see the output for a failed "invalidation test"? In the ones listed above, there is not much of visual output.

ranocha:
In SciML, we decided to skip CI for this (since we needed to update ~100 repos). I changed the commit message here to let CI run.
I don't have an example of a failing check.

ranocha:
This is how it looks like when you create invalidations: https://github.com/ranocha/InvalidationsTest.jl/pull/1

ranocha:
In my experience, fixing invalidations is a quite manual process, so you also need to experiment locally.
Please note that the numbers of invalidations are reported, too (in a previous step).

PR #191, opened 2022-09-09
Optimize `vector ⊗ vector` case

dkarrasch:
This is a major improvement for a narrow, but potentially interesting case: https://github.com/MichielStock/Kronecker.jl/issues/91.


dkarrasch:
@PythonNut Do you wanna take a look and see if it's useful for you?

Issue #193, opened 2022-10-10
Unnecessary allocation for mat-vec multiplication of composite map


felixhorger:
Hi, very useful package!
I work with huge vectors (~30Gb) and encountered a problem with composite maps.
`(A * B) * x` gives the same as `A * (B * x)` as expected, but the former allocates twice the space.
A and B are of type `FunctionMap` and have their own output arrays, e.g.  `A = LinearMap{T}(x -> zeros(42), 42)`.
In such a case, the output is copied to another allocated array, see this [function](https://github.com/JuliaLinearAlgebra/LinearMaps.jl/blob/bd10c6c74ada158d6b820694ccde21235ed01538/src/composition.jl#L202).
In my case that meant that around 50% of the time to compute `(A * B) * x` was spend on allocating memory or copying.
Is there any way to prevent this? A couple of my naive attempts to solve this failed.
In above example I could use the `ismutating` option, but my case is more complicated and that doesn't work anymore.

Cheers,
Felix

dkarrasch:
Hi @felixhorger! Thanks for the report. I believe the relevant piece of code for you is

https://github.com/JuliaLinearAlgebra/LinearMaps.jl/blob/bd10c6c74ada158d6b820694ccde21235ed01538/src/composition.jl#L176-L182

As you say, when you have a non-mutating `LinearMap`, then `_unsafe_mul!` essentially computes `B*x`, then copies the result to `source` (which was created when calling the function), and then computes the second factor `A*source`, which is only then copied to the final `y`. A quickfix could be to overload the "two-map" `_composite_mul!`. The issue is quite generic for outofplace `FunctionMap`s, because every call to `_unsafe_mul!` will create twice the necessary memory.

Perhaps a more sustainable solution is to make `FunctionMap` have a type parameter for whether it's an inplace-function or not. Then one could define a `mul!(y, A::CompositeMap{<:Any,Tuple{Vararg{FunctionMap{<:Any,false}}}}, x)` method that just computes `A[i] * x` consecutively. Or define a restricted `LinearMap` type that doesn't even define `mul!`, but only `*`.

Which other features do you need for your map, beyond map-vec-multiplication?

felixhorger:
Apologies for the late reply, a conference deadline kept me from pursuing this!
[...]

PR #194, opened 2022-10-12
Make `ismutable` a type-parameter of `FunctionMap`

dkarrasch:
This takes a small step towards what I think is wide-spread in the SciML/DiffEq world: making mutability a type-parameter. This allows to solve #193 by multiple dispatch, rather than querying a specific field of all factors/summands. The downside is that previous constructor calls are potentially not fully inferred, see the changed tests. If we advise users, however, to use the `FunctionMap` constructor explicitly and provide the mutability information, that is most likely statically known anyway, directly, then that shouldn't do much harm. In any case, the current design is such that the "old" behavior still works, it's just deprecated. What I find nice is that for out-of-place `FunctionMap`s, `A*B*C*D*x` first folds from the left, so that we first create a `CompositeMap` until we hit `x`, from where it folds from the right, without going through 3-arg `mul!`.


dkarrasch:
@JeffFessler I suspect you are using `FunctionMap`s heavily. Can you please try your packages with this PR branch and report back any new issues? I believe nothing should break, only some deprecation warnings should appear, but you never know.

src/LinearMaps.jl
```diff
@@ -18,8 +17,9 @@ using Base: require_one_based_indexing
 
 abstract type LinearMap{T} end
 
-const MapOrVecOrMat{T} = Union{LinearMap{T}, AbstractVecOrMat{T}}
-const MapOrMatrix{T} = Union{LinearMap{T}, AbstractMatrix{T}}
+const AbstractVecOrMatOrQ{T} = Union{AbstractVecOrMat{T}, AbstractQ{T}}
```
JeffFessler, line 20: I had never head of `AbstractQ` before so I searched it and it is a subtype of AbstractMatrix so why do we need this new union here?
https://github.com/JuliaLang/julia/blob/670190c3393e7a2193ad9e40cfa12ec643ceba75/stdlib/LinearAlgebra/src/qr.jl#L519

dkarrasch: Because of https://github.com/JuliaLang/julia/pull/46196. This is in preparation of that PR. It makes no difference on older Julia versions, but will on v1.10.

dkarrasch: I should perhaps split the two things in two separate PRs to not mix it up...

JeffFessler: Oh wow, I see.  Thanks for the explanation.  I am not looking forward to having to add more such `Union's` to support things that are kind of like a `AbstractMatrix` (in terms of multiplying) but not a subtype of it.  I've subscribed to the discussion of 46196 so I can see how that debate goes...  Now I'll try to see if this branch works for me.

dkarrasch:
Sorry I messed a few things up while splitting. Seems fixed now.

JeffFessler:
Ok, I did a test of `LinearMapsAA.jl` and it passed, but with only one deprecation warning so maybe that package does not actually use so many `FunctionMap`s.  So then I also tested `MIRT.jl`and again it passed with exactly one deprecation warning. Maybe that means that `@deprecate` warns only once ( So I guess that bodes well.  I suspect that many of my uses of FMs are spread out in many Literate examples that are not so easy to test but we'll find out in time...

I have not done much (if any) testing of PR branches of others so I want to describe the approach so that you can confirm if I did it properly.  I started Julia 1.8 in an empty directory, then:
```julia
] activate .
add LinearMaps#dk/iipfunctionmap
add LinearMapsAA
test LinearMapsAA
add MIRT
test MIRT
```

So LTGM overall, and I am sure that #193 will bite me too at some point so I am in favor of this direction!

dkarrasch:
Thank you very much for the detailed report! If you received deprecation warnings (no matter how many), then it grabbed the right branch. I'm glad that tests pass, too. I hope that `FunctionMap` construction is not done in hot loops, so the deprecation warnings should not yield a performance penalty. I guess this means that it is fine to release this with a minor version bump. For the future, I believe this will allow for some memory optimizations, even though we don't want to dive too deep in there.

fredrikekre:
A bit of a drive-by comment, but I personally find `FunctionMap{T,true}(...)` to be less readable compared to `FunctionMap{T}(..., ismutating=true)`. I tried to look at the code, but didn't see where the new parameter is used for dispatch. Does the parameter improve something external or is it just to get rid of `if`s internally?

dkarrasch:
Drive-by comments are always welcome! There is dispatch on `IIPFunctionMap` vs. `OOPFunctionMap` in linearcombinations.jl and composed.jl. And yes, it's used to catch cases where all `FunctionMap`s allocate the result array themselves. I don't have a strong preference, it just reminded me that this has been standard in the DiffEq world for a long time, but it could be solved equally well via branching internally I guess. Any thoughts?

JeffFessler:
> less readable compared to `FunctionMap{T}(..., ismutating=true)`

I tend to agree.  And the `kwarg` approach has the advantage of being amenable to splatting `kwargs...`
I prefer that approach for the `T=` argument too, but I probably am an outlier there.
I am talking about an outer constructor only; I realize the type itself and the inner constructor could have all sorts of parameters like `FunctionMap{T, F1, F2, iip}(...)`

dkarrasch:
Hm, another option could be to make `LinearMaps.ismutating` a "trait" for all `LinearMap` subtypes. That would allow asking a `LinearCombination`/`CompositeMap` whether all its maps are mutating, whereas the current design only catches cases where out-of-place action is due to `FunctionMap`s. But other, custom map types could have the same, and we would miss that. And we could even ask individual maps and do something with that information.

dkarrasch:
I realized we already have some kind of the necessary machinery: `MulStyle`, which so far has only `ThreeArg` and `FiveArg`, but which could be extended to `TwoArg`, meaning `A*x`. Reflecting in-place vs. out-of-place behaviour _reasonably_ via `MulStyle`, however, requires that this information is available in the type domain. So how about adding the type parameter `iip`, but not deprecate the old constructors? The only issue with that is that the `iip` parameter may not be inferred from the kwarg constructors. If, however, between map construction and extensive usage is a function barrier, that should be no problem. Thoughts?

JeffFessler:
I think it is fine not to have type inference for the kwarg constructors because (at least in my use cases) construction is done just once in an initialization phase and not in any tight loop.  Maintaining the old constructors sounds convenient to me.

test/linearcombination.jl
```diff
@@ -10,6 +11,18 @@ using LinearMaps: FiveArg, LinearMapTuple, LinearMapVector
...
+    @test Loop * v ≈ 3cumsum(v)
+    @test (@allocated Loop * v) <= 3alloc
+    Loop = @inferred CS + CS; Loop * v
+    @test (@allocated Loop * v) <= 2alloc
+    Lmix = @inferred CS + CS + CS!; Lmix * v
+    @test (@allocated Lmix * v) <= 3alloc
+    Lmix = @inferred CS + CS + CS!; Lmix * v
+    @test (@allocated Lmix * v) <= 3alloc
+    Lmix = @inferred CS! + (CS + CS); Lmix * v
+    @test (@allocated Lmix * v) <= 3alloc
```
dkarrasch, line 25: These tests are all sharp. They fail on the currently released version!

dkarrasch:
This is now new-feature-only, nothing breaks (except for the lack of complete type inference for the convenience constructor). Thanks @fredrikekre for retriggering some thought, thanks @JeffFessler for your comments and testing!

dkarrasch:
BTW, I changed the default values of `dest` and `source` in `_compositemul!` to `nothing`, and moved the allocation to the function body (if necessary). We don't use these temporary arrays internally, and any external user will take care of the right size and eltype themselves.

PR #198, opened 2022-12-15
Flag derivative WRT map itself as not implemented in chain rule

gaurav-arya:
There are definitely scenarios where there ought to be a gradient WRT some parameters of the linear map, so this PR marks its gradient as not implemented rather than as `NoTangent()` to try to avoid correctness issues. Here's an example:
```
function f(p)
           A = LinearMap([p])
           return sum(A * [p])
end
# function is a fancy p^2, so gradient should be 2p
```

Annoyingly, this doesn't actually fix anything yet, at least for `Zygote`, because `Zygote` treats `NotImplemented`'s just like zeros (https://github.com/FluxML/Zygote.jl/issues/1204). But it still seems like the right thing to do.

### DIFF
--- a/src/LinearMaps.jl
+++ b/src/LinearMaps.jl
@@ -11,7 +11,7 @@ using SparseArrays
 
 import Statistics: mean
 
-using ChainRulesCore: unthunk, NoTangent, @thunk
+using ChainRulesCore: unthunk, NoTangent, @thunk, @not_implemented
 import ChainRulesCore: rrule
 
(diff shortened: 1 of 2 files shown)


dkarrasch:
I'm not familiar with the chain rules stuff, but I trust that this is correct. Seems to be closer to the actual fact that the product in itself is differentiable w.r.t. map (in contrast to `NoTangent()`), but it's just not implemented.

PR #206, opened 2023-06-06
Handle ChainRulesCore via Pkg extension

oschulz:
Substantially reduces the load time of LinearMaps on Julia v1.9.

Before:

```julia
julia> @time import LinearMaps
  0.053604 seconds (73.57 k allocations: 4.871 MiB, 6.36% compilation time)
```

```julia
julia> @time_imports import LinearMaps, ChainRulesCore
      0.6 ms  Statistics
      0.2 ms  Compat
      0.2 ms  Compat → CompatLinearAlgebraExt
     24.7 ms  ChainRulesCore
     25.3 ms  LinearMaps
```

After:

```julia
julia> @time import LinearMaps
  0.037479 seconds (39.45 k allocations: 2.582 MiB, 12.74% compilation time)
```

```julia
julia> @time_imports import LinearMaps, ChainRulesCore
      0.5 ms  Statistics
     24.8 ms  LinearMaps
      0.2 ms  Compat
      0.1 ms  Compat → CompatLinearAlgebraExt
     30.9 ms  ChainRulesCore
      0.2 ms  LinearMaps → LinearMapsChainRulesCoreExt
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LinearMaps"
 uuid = "7a12625a-238d-50fd-b39a-03d52299707e"
-version = "3.10.0"
+version = "3.10.1"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"


dkarrasch:
Thank you very much! Do you understand the complaint by `Aqua.jl`? Do we need to move the additions to `Project.toml` to a different place? The load time gain is great. Looks like almost all the load time is spent on `ChainRulesCore`?

oschulz:
@dkarrasch The Aqua think I had hoped was handled now, but apparently not yet. It's just that Julia v1.6 formats Project.toml differently if there are weakdeps entries (which it doesn't know about) in it. So the Project.toml formatting tests has to be disabled on v1.6.

The invalidations failure seems unavoidable with Pkg extensions right now, see PainterQubits/Unitful.jl#652 .

oschulz:
Aqua tests should pass now, hopefully.

dkarrasch:
Could you please bump the patch number and then I'll release shortly after merging. Thanks again!

oschulz:
Version is bumped, thanks @dkarrasch !

PR #208, opened 2023-06-27
Handle dependencies via Pkg extensions

### LINKED ISSUE #207: Further reducing load time?
I recently created a very lightweight `LinearMap`-like type [`MatrixLikeOperator`](https://github.com/oschulz/AutoDiffOperators.jl/blob/main/src/matrix_like_operator.jl) for AutoDiffOperators.jl. AutoDiffOperators supports LinearMaps as an extension (and that's what I'll use in most "serious" use cases), but I wanted to keep load-time down and have a lightweight simple default linear map type.

While optimizing load time I noticed that many method invalidations originated from using `LinearAlgebra.Adjoint{<:MatrixLikeOperator}` to represent and dispatch on adjoints of the maps. So I added a custom adjoint type (like LinearMaps) and use a two-stage dispatch for `*` and `mul!`. In the end the load time went from over 30 ms down to 10 ms.

LinearMaps is probably pretty optimized already, but I wonder I we can bring load time down a bit more still, based on that? In the end, the number of `*` and `mul!` invalidations should be constant if we re-route to internal versions for different map types. Though LinearMaps does support a 5-arg `mul!` and my lightweight maps only do 3-arg `mul!`  - not sure ...

dkarrasch:
This handles the two dependencies `SparseArrays.jl` and `Statistics.jl` as weak dependencies. Indeed, the only reason LinearMaps.jl depended on these was to overload their functions. Otherwise, LinearMaps.jl has always been independent from these two. Moreover, users could not call, say, `sparse` without loading `SparseArrays.jl`, so this is a clear case. This brings package load time of pure `LinearMaps.jl` down to below 20 ms (from about 300 ms, if not 500 ms and more) on current Julia nightly with SparseArrays.jl out of the sysimage).

Resolves #207.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LinearMaps"
 uuid = "7a12625a-238d-50fd-b39a-03d52299707e"
-version = "3.10.1"
+version = "3.10.2"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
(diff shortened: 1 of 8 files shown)


dkarrasch:
@oschulz This should help your package, shouldn't it? It seems like your package doesn't depend on SparseArrays.jl, so you would avoid loading it, at least via LinearMaps.jl.

oschulz:
Nice, LinearMaps loads in 19 ms now. :-)

oschulz:
Are you tagging a new relase with this @dkarrasch ?

PR #210, opened 2023-07-02
Overload `LinearAlgebra.tr`

dkarrasch:
Some users expressed interest in the trace function, so here we go. In the generic case, this is, of course, expensive, but in some cases we may leverage the lazy structure and algebraic rules to compute the trace allocation-free, like in the (lazy) Kronecker product of two square matrices. Another included classic use case is (lazy) rank-one matrices.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LinearMaps"
 uuid = "7a12625a-238d-50fd-b39a-03d52299707e"
-version = "3.10.2"
+version = "3.11.0"
 
 [deps]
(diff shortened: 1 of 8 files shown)


JeffFessler:
The fallback code for general types might have been simpler if we allowed `A[i,i]` 🥲

dkarrasch:
True, but less efficient, because how would you reuse the canonical vector?

dkarrasch:
I wonder if we should have some kind of iterator for iterating all canonical vectors, though. Let me think about it...

JeffFessler:
Ah yes, fair point.

Issue #123, opened 2020-12-11
LinearMaps 3.0: Incompatibility with StaticArrays due to generic _cat


mfherbst:
I observed the issue when using LinearMaps and StaticArrays at the same time. Could be more general and effect other generic Array types as well, but I have not tested.

## Reproducer

Works fine:
```julia
using LinearMaps
using StaticArrays
rotations = [ones(3,3), ones(3,3)]
cat(rotations..., dims=3)
```

Not working any more with LinearMaps 3.0:
```julia
using LinearMaps
using StaticArrays
rotations = SMatrix{3,3,Float64}[ones(3,3), ones(3,3)]
cat(rotations..., dims=3)
```

Stacktrace:
```
ERROR: TypeError: in typeassert, expected Tuple{Int64,Int64}, got a value of type Int64
Stacktrace:
 [1] _cat(::Int64, ::SArray{Tuple{3,3},Float64,2,9}, ::Vararg{SArray{Tuple{3,3},Float64,2,9},N} where N) at /packages/LinearMaps/OwJoG/src/blockmap.jl:455
 [2] cat(::SArray{Tuple{3,3},Float64,2,9}, ::Vararg{SArray{Tuple{3,3},Float64,2,9},N} where N; dims::Int64) at ./abstractarray.jl:1654
 [3] top-level scope at REPL[19]:1
```

mfherbst:
Thanks for your quick reply ... yes patching these sort of things in Base is always a bit tricky.

PR #216, opened 2023-09-26
latency: fix cat invalidations

johnnychen94:
This patch removes the invalidation on cat and thus reduces the OneHotArrays loading time from 4.5s to 0.5s (the normal status)

`using OneHotArrays` alone in a new Julia process needs about 0.57s. However, `using LinearMaps` before that makes the loading time of OneHotArrays to about 5s.

```julia
julia> @time using OneHotArrays # v0.2.4
  0.569807 seconds (539.89 k allocations: 36.725 MiB, 4.48% compilation time)
```

```julia
# in a new Julia process
# LinearMaps master branch
julia> @time using LinearMaps
  0.023610 seconds (6.77 k allocations: 501.328 KiB)

julia> @time using OneHotArrays
  5.038436 seconds (13.72 M allocations: 514.807 MiB, 2.32% gc time, 0.51% compilation time)
```

With SnoopCompile I've noticed LinearMaps invalidates the `cat` methods and [OneHotArrays extends the `cat` methods](https://github.com/FluxML/OneHotArrays.jl/blob/f10062381af720df4e661899403f3a63aea6f2ad/src/array.jl#L112-L120):

```julia
julia> using SnoopCompileCore

julia> invalidations = @snoopr begin
           using LinearMaps
       end
174-element Vector{Any}:
   MethodInstance for Core.kwcall(::@NamedTuple{dims::Val{1}}, ::typeof(cat), ::Vector{Base.PkgId}, ::Vararg{Any})
   "invalidate_mt_cache"
   ...
```

This patch tweaks the `cat` method extension by dispatching on `_cat` instead of `cat`, and somehow it does work:

```julia
julia> @time using LinearMaps
  0.012467 seconds (11.94 k allocations: 741.312 KiB)

julia> @time using OneHotArrays # 🎉 back to normal
  0.576323 seconds (539.17 k allocations: 36.551 MiB, 4.37% compilation time)
```

---

P.S. I figure this is related to https://github.com/JuliaLang/julia/pull/45028 (because the `Core.kwcall` hint in the invalidations list) but I don't know why this works. Maybe @aviatesk or @timholy knows the magic behind?

Commenting out the `Base.cat` method in above link in OneHotArrays also reduces the latency, but I figure it's better to do it here since OneHotArrays looks more innocent while this packages code generates many cat methods.
FWIW, if we reduce the number `for k in 1:8` to something like `for k in 1:4` the latency won't be that severe, too. As an experiment:

```diff
diff --git a/src/blockmap.jl b/src/blockmap.jl
index 72c2bb9..49ba024 100644
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -497,7 +497,7 @@ BlockDiagonalMap(maps::LinearMap...) =

 # since the below methods are more specific than the Base method,
 # they would redefine Base/SparseArrays behavior
-for k in 1:8 # is 8 sufficient?
+for k in 1:16 # is 8 sufficient?
     Is = ntuple(n->:($(Symbol(:A, n))::AbstractVecOrMatOrQ), Val(k-1))
     # yields (:A1, :A2, :A3, ..., :A(k-1))
     L = :($(Symbol(:A, k))::LinearMap)
```

This makes `using OneHotArrays` takes forever.

### DIFF
--- a/src/blockmap.jl
+++ b/src/blockmap.jl
@@ -505,13 +505,28 @@ for k in 1:8 # is 8 sufficient?
     mapargs = ntuple(n ->:($(Symbol(:A, n))), Val(k-1))
     # yields (:LinearMap(A1), :LinearMap(A2), ..., :LinearMap(A(k-1)))
 
-    @eval function Base.cat($(Is...), $L, As::MapOrVecOrMat...; dims::Dims{2})
-        if dims == (1,2)
-            return BlockDiagonalMap(convert_to_lmaps($(mapargs...))...,
-                                    $(Symbol(:A, k)),
-                                    convert_to_lmaps(As...)...)
-        else
-            throw(ArgumentError("dims keyword in cat of LinearMaps must be (1,2)"))
+    @static if VERSION >= v"1.8"
+        # Dispatching on `cat` makes compiler hard to infer types and causes invalidations
+        # after https://github.com/JuliaLang/julia/pull/45028
+        # Here we instead dispatch on _cat
+        @eval function Base._cat(dims, $(Is...), $L, As...)
+            if dims == (1,2)
+                return BlockDiagonalMap(convert_to_lmaps($(mapargs...))...,
+                                        $(Symbol(:A, k)),
+                                        convert_to_lmaps(As...)...)
+            else
+                throw(ArgumentError("dims keyword in cat of LinearMaps must be (1,2)"))
+            end
+        end
+    else
+        @eval function Base.cat($(Is...), $L, As...; dims::Dims{2})
+            if dims == (1,2)
+                return BlockDiagonalMap(convert_to_lmaps($(mapargs...))...,
+                                        $(Symbol(:A, k)),
+                                        convert_to_lmaps(As...)...)
+            else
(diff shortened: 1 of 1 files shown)


dkarrasch:
First I was afraid this would bring back #123 (see #124 for the offending method signature), but this works since you keep the method signature as is. Unfortunately, on my machine, this PR increases loading time from 0.05 seconds to about 7 seconds on v1.9!

johnnychen94:
Looks like this isn't the right fix. I'll try to see if there's any better solution then.

dkarrasch:
Interestingly, this seems to work well for Julia v1.10+. Unfortunately, load time of `OneHotArrays.jl` doubles from v1.10 to v1.11 and master even without any interference by `LinearMaps.jl`. But this PR increases load time of `LinearMaps.jl` by just a little bit, and decreases load time of `OneHotArrays.jl` after `LinearMaps.jl` dramatically (factor 8, roughly).

dkarrasch:
I like this, and I think we should push this even further: ride the generic call chain and catch somewhere much deeper, where there is almost no method competition. Doesn't work yet, but pushed to continue working on it from work tomorrow.

PR #217, opened 2023-10-04
Don't load packages in extension other than parent

dkarrasch:
Apparently, this is how it should be done. Non-functional change in any case.


fredrikekre:
Whats wrong with the current code?

dkarrasch:
TIL that this is how it should be done: https://github.com/JuliaStats/LogExpFunctions.jl/pull/63 Or was that only true at some intermediate point in development?

dkarrasch:
Seems like this is not necessary (anymore).

PR #221, opened 2024-01-25
Always use similar as fallback when resize throws in applying composite maps

krcools:
When using subtypes of `AbstractVector` that do not support `resize!` (such as `BlockArrays.PseudoBlockVector`), the memory optimisation in the `_unsafe_mul!` for composite maps result in errors.

I have changed `_resize` to allways fall back to memory allocation via `similar`, whereas previously this only happened when the argument was a `SharedArrays.SharedVector`.


dkarrasch:
LGTM. Do you have a quick test case at hand? I believe we have BlockArrays.jl in our test dependencies anyway.

krcools:
I have added a test with a triple product of LinearMaps over BlockArrays in nontradaxes.jl

dkarrasch:
It turns out that it requires four maps to hit that `_resize` path. 😄

dkarrasch:
Thanks! That's a clear improvement, since otherwise there was no way to get this to work, right?

krcools:
> It turns out that it requires four maps to hit that `_resize` path. 😄

Thank you, I missed that the triple product would return before `_resize` got hit.

> Thanks! That's a clear improvement, since otherwise there was no way to get this to work, right?

I think so, that was my motivation to submit the PR. A better solution is probably to implement `_resize` only for `Vector`, as I understand from the Julia coding recommendations that exception handling is rather expensive...

PR #248, opened 2026-05-20
Clarify docs regarding `FiveArg()` trait

### LINKED ISSUE #236: Documentation of FiveArg()
If we set the trait FiveArg(), is it also necessary to implement the three argument `mul!`? Something like:
```
LinearMaps.MulStyle(A::MyMap) = FiveArg()

function LinearMaps._unsafe_mul!(y, A::MyMap, x::AbstractVector, α, β)
    ...
end

LinearMaps._unsafe_mul!(y, A::MyMap, x::AbstractVector) = LinearMaps._unsafe_mul!(y, A, x, 1, 0)
```

If so, it would be useful to note this in the "Custom Maps" section of the documentation.
If you don't need to do this then I have some other issue... :)

dkarrasch:
Closes #236.

### DIFF
--- a/docs/src/custom.jl
+++ b/docs/src/custom.jl
@@ -118,6 +118,11 @@ end
 
 # There you go, the allocation is gone and the computation time is significantly reduced.
 
+# !!! note
+#     Unlike in the standard library `LinearAlgebra.jl`, there is no automatic forwarding
+#     of 3-arg `mul!` methods to those with 5 arguments, so you need to define the basic
+#     3-arg `_unsafe_mul!` method in any case.
+
 # ## Adjoints and transposes
 
 # Generically, taking the transpose (or the adjoint) of a (real, resp.) map wraps the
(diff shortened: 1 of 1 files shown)


dkarrasch:
@ruairidhs Does this help?

@JeffFessler Could you please take a look at the language?

docs/src/custom.jl
```diff
@@ -117,6 +118,11 @@ end
 
 # There you go, the allocation is gone and the computation time is significantly reduced.
 
+# !!! note
+#     Unlike in the standard library `LinearAlgebra.jl`, there is no automatic forwarding
```
JeffFessler, line 122: all of the updated description looks good to me, except that it might be helpful to explain *why* there is no automatic forwarding here?

dkarrasch: Thanks for your feedback. I added a note.

